Issue #49, opened 2016-12-16
layer: Xattr unpack code is broken on Fedora/RHEL


runcom:
just have a busybox image copied with skopeo to OCI format, then:
```sh
$ umoci unpack --image busybox bundle
INFO[0000] parsed mappings                               map.gid=[] map.uid=[]
FATA[0000] create runtime bundle: chown rootfs: lchown bundle/rootfs: operation not permitted

$ sudo $GOPATH/bin/umoci unpack --image busybox bundle
INFO[0000] parsed mappings                               map.gid=[] map.uid=[]
INFO[0000] unpack manifest: unpacking layer sha256:56bec22e355981d8ba0878c6c2f23b21f422f30ab0aba188b54f1ffeff59c190  diffid="sha256:e88b3f82283bc59d5e0df427c824e9f95557e661fcb0ea15fb0fb6f97760f9d9"
FATA[0000] create runtime bundle: unpack layer: unpack entry: bin: apply hdr metadata: clear xattr metadata: /home/amurdaca/go/src/github.com/docker/containerd/bundle/rootfs/bin: lclearxattrs: lremovexattr(/home/amurdaca/go/src/github.com/docker/containerd/bundle/rootfs/bin, security.selinux): permission denied
```

runcom:
if I add `--rootless` then everything appears to work

cyphar:
Yup. I had a comment about "this probably breaks on RedHat". I'm not super familiar about `Xattrs` and what I should be doing here.

What I currently do is this:

1. Remove all xattrs that exist on a file if it already exists.
2. Set all xattrs that are specified in the header.

This is clearly not correct, so what should I be doing? Should I be modifying `xattrs` in place? What should I do about xattrs that are set on files that were not present in the image?

runcom:
> Set all xattrs that are specified in the header.

clearly my command fails there right?

cyphar:
> clearly my command fails there right?

No, it's failing when clearing. You can pass `--debug` to get the full backtrace from `pkg/errors`. But the wrapped error is "clear xattr metadata".

```
create runtime bundle: unpack layer: unpack entry: bin: apply hdr metadata: clear xattr metadata: /home/amurdaca/go/src/github.com/docker/containerd/bundle/rootfs/bin: lclearxattrs: lremovexattr(/home/amurdaca/go/src/github.com/docker/containerd/bundle/rootfs/bin, security.selinux): permission denied
```

[...]

PR #51, opened 2016-12-16
pkg: xattr: ignore EPERM in Lclearxattrs

cyphar:
On RHEL/Fedora this fixes issues under SELinux where trying to remove a
"security.selinux" label is a capital-B bad idea. It also might be
necessary to handle blacklisting "security.*" xattrs when we generate
archives.

Fixes: cyphar/umoci#49

### DIFF
--- a/pkg/system/xattr_linux.go
+++ b/pkg/system/xattr_linux.go
@@ -19,6 +19,7 @@ package system
 
 import (
 	"bytes"
+	"os"
 	"syscall"
 	"unsafe"
 
@@ -52,6 +53,11 @@ func Llistxattr(path string) ([]string, error) {
 
 	var xattrs []string
 	for _, name := range bytes.Split(buffer, []byte{'\x00'}) {
+		// "" is not a valid xattr (weirdly you get ERANGE -- not EINVAL -- if
+		// you try to touch it). So just skip it.
+		if len(name) == 0 {
+			continue
+		}
 		xattrs = append(xattrs, string(name))
 	}
 	return xattrs, nil
@@ -121,11 +127,16 @@ func Lgetxattr(path string, name string) ([]byte, error) {
 func Lclearxattrs(path string) error {
 	names, err := Llistxattr(path)
 	if err != nil {
-		return errors.Wrap(err, "lclearxattrs")
+		return errors.Wrap(err, "lclearxattrs: get list")
 	}
 	for _, name := range names {
 		if err := Lremovexattr(path, name); err != nil {
-			return errors.Wrap(err, "lclearxattrs")
+			// Ignore permission errors, because hitting a permission error
+			// means that it's a security.* xattr label or something similar.
+			if os.IsPermission(errors.Cause(err)) {
+				continue
+			}
+			return errors.Wrap(err, "lclearxattrs: remove xattr")
 		}
 	}
 	return nil


cyphar:
LGTM

cyphar:
LGTM

cyphar:
LGTM

cyphar:
LGTM

cyphar:
LGTM

runcom:
@cyphar great, today I also woke up thinking of https://github.com/projectatomic/docker/commit/48cc4ab60a5fa2b70b3dbfff05179fbfa8771fa4 :smile: 

This fixes #49 when running with `sudo`:
```sh
10:38:03 [github.com/cyphar/umoci] ‹pr/origin/51*› sudo ./umoci unpack --image busybox bundle
INFO[0000] parsed mappings                               map.gid=[] map.uid=[]
INFO[0000] unpack manifest: unpacking layer sha256:56bec22e355981d8ba0878c6c2f23b21f422f30ab0aba188b54f1ffeff59c190  diffid="sha256:e88b3f82283bc59d5e0df427c824e9f95557e661fcb0ea15fb0fb6f97760f9d9"
INFO[0000] unpack manifest: unpacking config             config="sha256:e02e811dd08fd49e7f6032625495118e63f597eb150403d02e3238af1df240ba"
```
However (not sure it's expected), unpacking w/o `sudo` or `--rootless` still fails:
```sh
10:39:27 [github.com/cyphar/umoci] ‹pr/origin/51*› ./umoci unpack --image busybox bundle
INFO[0000] parsed mappings                               map.gid=[] map.uid=[]
FATA[0000] create runtime bundle: chown rootfs: lchown bundle/rootfs: operation not permitted
```

cyphar:
@runcom 

> However (not sure it's expected), unpacking w/o sudo or --rootless still fails:

Yes, you have to _explicitly specify_ that you want `--rootless` because it does a bunch of tricks (that I'm quite proud of) that you wouldn't want to be done by default.

cyphar:
LGTM. Thanks @runcom for checking. :smile_cat:

runcom:
> Yes, you have to explicitly specify that you want --rootless because it does a bunch of tricks (that I'm quite proud of) that you wouldn't want to be done by default.

I see 
```
119     // Make sure that the owner is correct.
120     rootUID, err := idtools.ToHost(0, opt.UIDMappings)
121     if err != nil {
122         return errors.Wrap(err, "ensure rootuid has mapping")
123     }
124     rootGID, err := idtools.ToHost(0, opt.GIDMappings)
125     if err != nil {
126         return errors.Wrap(err, "ensure rootgid has mapping")
127     }
128     if err := os.Lchown(rootfsPath, rootUID, rootGID); err != nil {
129         return errors.Wrap(err, "chown rootfs")
130     }
```
Shouldn't this own the rootfs to the calling $USER uid/gid instead of `0` by default when running w/o sudo or --rootless? (just a silly question)

cyphar:
In case you're interested, I'm working on proper `xattr` repacking as well in #54 -- so you can actually add xattrs. There was a bug in go-mtree that I just fixed (should work _soon_).

runcom:
> Yes, you have to explicitly specify that you want --rootless because it does a bunch of tricks (that I'm quite proud of) that you wouldn't want to be done by default.

alright, I'm blind, gotcha, thanks @cyphar - maybe worth adding a "try with --rootless" to the error

cyphar:
@runcom Yeah, I really need to improve the amount of crap I'm outputting -- you get debug output by default and no helpful hints about what to do.

Maybe I should make it so that if an error bubbles up to `cmd/umoci/main` and `os.IsPermission(errors.Cause(err))` then I output that message? I'll see. :P

PR #54, opened 2016-12-17
oci: layer: add xattrs to new layers

cyphar:
This implements xattr handling in tar_generate, as well as some other
cleanups.

By necessity we have to ignore some xattrs (such as security.selinux),
but hopefully this is a rare case of things we need to ignore -- I don't
really want to have to maintain a set of xattrs that are and are not
safe to handle.

Fixes: cyphar/umoci#52

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,18 +21,21 @@ MAINTAINER "Aleksa Sarai <asarai@suse.com>"
 RUN zypper ar -f -p 10 -g obs://Virtualization:containers obs-vc && \
     zypper ar -f -p 10 -g obs://devel:languages:go obs-dlg && \
     zypper ar -f -p 15 -g obs://home:cyphar obs-home-cyphar && \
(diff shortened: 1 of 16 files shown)


cyphar:
For some reason this doesn't work with `user.XYZ` xattrs. My guess is that it's because `tar.Xattr` is not doing a good job.

cyphar:
Currently `go-mtree` won't correctly detect the addition of an `xattr` because of how `Compare` implements the checking. It used to work but was broken in cleanups (because I never used `xattr` and the tests didn't check for it). See vbatts/go-mtree#107

cyphar:
Tests are failing, not sure why. In particular, it looks like the `user.*` xattrs aren't being added for some reason. On the one hand, AUFS has some fun interactions with extended attributes, on the other `xattr` and `unpack` aren't spitting out errors!

cyphar:
LGTM

PR #56, opened 2016-12-17
fseval: add Mknod

cyphar:
This also includes the change to unpriv, so that we can do unprivileged
Mknod creation. Currently I haven't put the hacks done for devices
because that code is quite frankly pretty dark.


cyphar:
LGTM

PR #59, opened 2016-12-17
hack: switch from glide to vendor.sh

cyphar:
Purely for my sanity.


cyphar:
LGTM

Issue #48, opened 2016-12-16
handle removal of Memory, MemorySwap and CPUShares


cyphar:
These never made sense in the spec, and they've been recently removed so we should drop all mention of them from everywhere. opencontainers/image-spec#495

PR #62, opened 2016-12-18
*: drop support for resource configs

cyphar:
This includes the removal of Config.{Memory{,Swap},CPUShares}, which is
a long time coming, considering that it never made any sense.

Fixes: #48


cyphar:
LGTM

PR #63, opened 2016-12-19
gc: clean internal directories

### LINKED ISSUE #17: gc: clean up tmpdirs with dirEngine
This is going to be a bit of a layer violation, so I'll have to think about it. But basically we need to have a way to remove all of the temporary directories that have been left behind inside an OCI image.

cyphar:
This adds support for cleaning of dead temporary directories inside
an OCI layout directory. It's all implemented using locking.

Fixes #17 


cyphar:
LGTM.

PR #64, opened 2016-12-19
Makefile: fix static compilation

cyphar:
The old setup didn't actually statically compile umoci, so we have to
add CGO_ENABLED=0 in order to make the damn thing static. Also,
installing dependencies is a bad idea.


cyphar:
LGTM.

PR #65, opened 2016-12-20
*: make unit tests more %check-friendly

cyphar:
This is necessary for %check-style testing inside the openSUSE RPMs, so
that we can make sure that everything works (or appears to work) when
releasing packages. This is done by making local-unit-tests work as
non-root.

It also increases unit test coverage for the rootless case.


cyphar:
LGTM.

PR #68, opened 2016-12-29
*: test coverage improvements

### LINKED ISSUE #67: test: increase coverage
I'd prefer if we can hit a unit test coverage of ~80%. Currently it's not enough testing IMO. The problem is that there's a lot of error paths that we will not be able to hit.

cyphar:
This includes a lot of improvements to test coverage. In addition to adding
more tests, the main cool thing is that we now generate coverage reports that
cover **all** of the testing done in `umoci`, so you can get a sense for how
tested the whole codebase is. Currently I'm aiming to get ~80% test coverage.

Fixes #67 


cyphar:
LGTM.

PR #70, opened 2016-12-30
third_party: user: fix 32-bit build

cyphar:
The test code apparently breaks on i596, fix it by not using strconv
anymore.


cyphar:
LGTM

PR #73, opened 2017-02-06
*: clean up logging

### LINKED ISSUE #71: reduce amount of redundant output
Currently all of our libraries and commandline wrappers provide way too much information about each step of extraction -- the user doesn't really care about all of the gory details about extraction and generation.

cyphar:
apex/log has much more useful handlers (formatters) for CLI tools. In
partciular, the "cli" handler is much better suited to command-line
tooling than any formatter available for logrus.

Add custom logging levels, defaulting to the logging level of warning.
This makes the output much less spammy.

Fixes #71 


cmd/umoci/main.go
```diff
@@ -65,14 +66,34 @@ func main() {
...
 		},
 	}
 
 	app.Before = func(ctx *cli.Context) error {
-		if ctx.GlobalBool("debug") {
-			logrus.SetLevel(logrus.DebugLevel)
+		log.SetHandler(logcli.New(os.Stderr))
+
+		if ctx.GlobalBool("verbose") {
+			if ctx.GlobalIsSet("log") {
```
runcom: You may want to no-op on log=info?

runcom:
Looks good!

cmd/umoci/main.go
cyphar: I'd prefer to make it so people don't mix the two at all. But thanks for the review.

cyphar:
Thanks @runcom!

cyphar:
LGTM

PR #78, opened 2017-02-08
cmd/umoci: tag: fix tag clobbering

cyphar:
Previously, tags would always be clobbered, but this logic was subtly
broken for "umoci tag". Fix this by matching the other PutReference
usages.

Ensure that we don't get a regression for "umoci tag" clobbering.

Fixes: 8b73b49 ("*: make logging [log.Info] much more user-friendly")


cyphar:
LGTM :+1::shipit:

PR #79, opened 2017-02-08
vendor: update to opencontainers/image-spec@v1.0.0-rc4

cyphar:
This also required adding the following other libraries:
 * opencontainers/go-digest@v1.0.0-rc0

The purpose of this update is to pull in the necessary MediaType removal
and similar changes, to ensure that we keep up to date with the changes
upstream (since skopeo has updated before us).


cyphar:
LGTM

PR #89, opened 2017-02-18
oci: cas: *ispec.Descriptor -> ispec.Descriptor

cyphar:
The previous API didn't make much sense, and was quite confusing when
actually trying to use it (nil would always cause errors, and the
intention of the call is not to modify the arguments).

This is part of the preparation of putting oci/cas into an OCI
repository.


cyphar:
LGTM

PR #90, opened 2017-02-18
oci/*: major split ups

cyphar:
In preparation for migrating most of `oci/*` to `image-tools` or some such
repo, we need to split up several packages so that they are less of a
conglomeration and more of a useful set of packages that implement the minimal
functionality so other packages can build on top of them.

This changeset also includes quite a few refactorings to make certain
interfaces more generic, in anticipation for the requirements that
`image-tools` needs to fulfill using these libraries.

Things that need to be done:

* [x] Split conversion (`image` -> `runtime`) from generation (`image`).
* [x] Refactor GC by implementing a generic `Walk` utility.
* [x] Split out `gc` and `walk` from `oci/cas`.

Possibilities:

* [x] Instead of implementing all of the wrappers around `cas.Engine` as functions that take it as an argument, implement some wrapping structure that provides extra methods (but still provides the old ones) -- like a derived type.

* [x] Split `oci/cas/dir.go` into some sort of "drivers"-style code so that multiple drivers can more easily be added later. To be honest, this isn't something we need right now -- and the code is already separate enough that you could split it by just moving the file (even the tests would continue to work).


cyphar:
NOTE: For some reason, imports like `github.com/openSUSE/umoci/oci/cas/gc` appear to get muddled and are imported as `cas`. Not sure why -- this needs to be fixed before this is merged. Oddly, you can _force_ the import:

```
import blob "github.com/openSUSE/umoci/oci/cas/blob"
```

wking:
On Sat, Feb 18, 2017 at 01:01:00PM -0800, Aleksa Sarai wrote:
>   * oci: oci/generate -> oci/spec/generate

I know runtime-tools uses ‘Spec’ for the config Go type, but I'd
recommend using ‘config’ instead of ‘spec’.  Lots of things are
spec'ed, but what you're generating here is an image-spec
configuration.  For more on this, see
opencontainers/runtime-tools#266.

wking:
On Sat, Feb 18, 2017 at 01:01:00PM -0800, Aleksa Sarai wrote:
>   * .: remove LICENSE

From the commit message for e764c54:

> It confuses GitHub's automated license detection.

Because the content you have there is the boilerplate, not the license
itself.  More on this exact issue in benbalter/licensee#103.  What you
should have there is the Apache license itself (for an example, see
opencontainers/project-template).

> The header in every file describing the license is sufficient.

I don't think that's true.  From the Apache license itself [1]:

> 4. Redistribution. You may reproduce and distribute copies of the
>    Work or Derivative Works thereof in any medium, with or without
>    modifications, and in Source or Object form, provided that You
>    meet the following conditions:
>
> a. You must give any other recipients of the Work or Derivative
>    Works a copy of this License; and

[1]: https://www.apache.org/licenses/LICENSE-2.0.html

cyphar:
@wking I already have a full copy of the license in COPYING. Not to mention that the header itself tells you where you can get a copy of the license. This is a

cyphar:
@wking I already have a full copy of the license in COPYING. Not to mention that the header itself tells you where you can get a copy of the license. This is _really_ not the point of this PR...

cyphar:
@wking I already have a full copy of the license in COPYING. Not to mention that the header itself tells you where you can get a copy of the license. This is _really_ not the point of this PR. It'd be great if you didn't derail a PR like this with license discussion.

wking:
On Sun, Feb 19, 2017 at 01:04:55PM -0800, Aleksa Sarai wrote:
> @wking I already have a full copy of the license in COPYING.

That should be all you need then.  No need to have a separate LICENSE
with just the boilerplate.  Removing it (like you're doing in this PR)
should un-confuse Licensee by avoiding benbalter/licensee#103.

> This is _really_ not the point of this PR. 

I'd missed the presense of a separate COPYING (sorry about that).
Maybe pull e764c549 into its own PR?  I think the change it's making
is good, although I'd prefer the commit message to be closer to:

  LICENSE: Remove boilerplate file (full license still in COPYING)

  It confuses GitHub's automated license detection
  (benbalter/licensee#103).  And with the full license in COPYING,
  there's no reason to have a separate file with just the boilerplate.

cyphar:
LGTM

PR #100, opened 2017-03-01
Install target

### LINKED ISSUE #99: build from source
I was wondering if there are instructions floating around on how to build umoci from a source clone ? I'm working with Open Containers in my yocto based container OS and I was about to write a similar tool to umoci when I stumbled here.

I'm already using umoci's dependencies as part of my work, so I have them packaged and ready to go .. but I've thus far been able to build oci without getting errors. I'm more of a casual go hacker, so I'm not having any luck figuring out what I've done wrong.

I'd like to build it by hand on my development system, before I start transforming that build into a bitbake recipe. I've arranged for all the source projects to be available, etc, just as I always do when integrating these sorts of packages.

So that goes back to my first question, If someone clone's umoci, what would you expect to be the steps to build it directly ?

erikh:
Here's a few makefile targets/refactors to support a `make install` target. PTAL.

Fixes #99

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,18 +30,30 @@ UMOCI_IMAGE := umoci_dev
 VERSION := $(shell cat ./VERSION)
 COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
 COMMIT := $(if $(shell git status --porcelain --untracked-files=no),"${COMMIT_NO}-dirty","${COMMIT_NO}")
+CMD := ./cmd/umoci
+
+DYN_BUILD_FLAGS := -ldflags "-s -w -X main.gitCommit=${COMMIT} -X main.version=${VERSION}" -tags "$(BUILDTAGS)"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -30,18 +30,30 @@ UMOCI_IMAGE := umoci_dev
...
+CMD := ./cmd/umoci
+
+DYN_BUILD_FLAGS := -ldflags "-s -w -X main.gitCommit=${COMMIT} -X main.version=${VERSION}" -tags "$(BUILDTAGS)"
+STATIC_BUILD_FLAGS := -ldflags "-s -w -extldflags '-static' -X main.gitCommit=${COMMIT} -X main.version=${VERSION}" -tags "$(BUILDTAGS)"
 
 .DEFAULT: umoci
 
-GO_SRC =  $(shell find . -name \*.go)
+GO_SRC = $(shell find . -name \*.go)
+.PHONY: umoci umoci.static umoci.cover install install.static
```
cyphar: Not sure that `umoci` should be a `.PHONY` target.

cyphar:
LGTM.

cyphar:
:100:

erikh:
\o/

PR #102, opened 2017-03-21
packaging: move to contrib/pkg

cyphar:
I don't really like top-level directories that are not source code
outside of contrib/ or docs/. This cleans up 5f388e3033a3 ("packaging:
add packaging helper to automate submissions to obs").


cyphar:
LGTM

PR #103, opened 2017-03-21
contrib: pkg: fix make_spec version

cyphar:
The version parsing was broken in e4f29ed51033 ("packaging: move to
contrib/pkg"). Also add an entry to the CHANGELOG.

It appears this broke openSUSE's build setup. Since my local developer
environment no longer relies on super-dodgy symlinks I don't mind being
explicit.


cyphar:
LGTM.

cyphar:
LGTM.

PR #106, opened 2017-04-04
test: clean up bundles in tests

### LINKED ISSUE #105: ci: travis failing because of insufficient cleanup
It looks like the Fedora image is so big that it causes `ENOSPC` after you run enough tests. Specifically it seems the cleanup sections aren't handling this properly (everything should be removed after each test runs).

[failure](https://travis-ci.org/openSUSE/umoci/jobs/217637435#L1776)

cyphar:
Previously we dropped bundle cleanups because rm(1) is not sophisticated
enough to know how to remove rootless rootfs-es. But this causes issues
with Travis (which has a cap on root filesystem size) so hack around the
rm(1) issues with a chmod 0777 -- which hopefully should handle all of
the important cases.

We have to track everything created with setup_testdir in a separate
manifest file because it turns out that global variables are hard to
modify in bash.

Closes #105 

### DIFF
--- a/test/config.bats
+++ b/test/config.bats
@@ -21,12 +21,13 @@ function setup() {
 }
 
 function teardown() {
+	teardown_tmpdirs
 	teardown_image
 }
 
 @test "umoci config" {
-	BUNDLE_A="$(setup_bundle)"
-	BUNDLE_B="$(setup_bundle)"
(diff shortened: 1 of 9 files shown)


cyphar:
LGTM.

PR #107, opened 2017-04-05
cmd: add support for --clear=config.{cmd,entrypoint}

cyphar:
While the current UX for --config.cmd and --config.entrypoint seems odd
(since they aren't treated atomically, which doesn't make any sense).
But since we're currently treating them like ENV we need to provide a
way to explicitly clear them (even though we clear the old value when
creating a new one).

Reported-by: David Cassany <dcassany@suse.com>


davidcassany:
Looks good to me, thanks! However I could not try it, apparently I do not have the appropriate tool chain for compiling it 😛

cyphar:
You shouldn't need anything more than `go`. I mean, you could create a branch of the `umoci` OBS project, but that's probably a bit too much work to test a ~20 line change.

cyphar:
LGTM.

PR #108, opened 2017-04-06
README: reuse the variable for install

vbatts:
Plus for those copying and pasting, if their `$GOPATH` is not `$HOME`
then they will not have to edit that line before pasting.

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>


cyphar:
LGTM

PR #109, opened 2017-04-07
unpack: don't set AdditionalGids in rootless mode

cyphar:
We can't join any additional groups as an unprivileged user, so
retaining these groups will always cause failures.

I found this while working on cyphar/orca-build#6.


cyphar:
LGTM.

cyphar:
LGTM

cyphar:
LGTM

Issue #21, opened 2016-11-06
config: implement raw-config


cyphar:
It would be nice if we could have a `--dry-run` flag that allows a user to take an input template (from anywhere) and outputs the modified config. If we correctly break out `mutateConfig` then this should actually be entirely doable by adding a new subcommand (`raw-config` or something).

cyphar:
It looks like this is needed for cyphar/orca#7 (in order for the build to be even remotely efficient). So the only real question is how we should handle things like `Config.User` without a rootfs (from a UX perspective).

cyphar:
Done by #110.

PR #110, opened 2017-04-09
cmd: umoci: add umoci-raw-runtime-config(1)

cyphar:
Add a new command to allows users that have to parse the generated
config.json often much more efficient (by allowing them to skip the
layer unpacking steps).

Since the output of this command is not quite obvious unless you know
what it's doing, hide it behind a "raw" subcommand that should scare
away users that don't know enough about the OCI specs.

Implements #21


cyphar:
LGTM.

PR #127, opened 2017-06-03
repack: do not include Volume paths in new layers

cyphar:
This is (unfortunately) not mandated in [the][1] [specification][2], but in
order to avoid accidentally spilling private information into published
layers (which is one use of Volumes) we must ignore all layers included
in Config.Volumes.

In future we should also add some flags or alernative ways of masking
paths.

[1]: opencontainers/image-spec#496
[2]: opencontainers/image-spec#504

### DIFF
--- a/cmd/umoci/repack.go
+++ b/cmd/umoci/repack.go
@@ -31,6 +31,7 @@ import (
 	"github.com/openSUSE/umoci/oci/casext"
 	igen "github.com/openSUSE/umoci/oci/config/generate"
 	"github.com/openSUSE/umoci/oci/layer"
+	"github.com/openSUSE/umoci/pkg/mtreefilter"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
@@ -64,6 +65,17 @@ manifest and configuration information uses the new diff atop the old manifest.`
 
 	// repack creates a new image, with a given tag.
 	Category: "image",
+
+	Flags: []cli.Flag{
+		cli.StringSliceFlag{
+			Name:  "mask-path",
+			Usage: "set of path prefixes in which deltas will be ignored when generating new layers",
+		},
+		cli.BoolFlag{
+			Name:  "no-mask-volumes",
+			Usage: "do not add the Config.Volumes of the image to the set of masked paths",
+		},
+	},
 
 	Action: repack,
 
@@ -155,6 +167,19 @@ func repack(ctx *cli.Context) error {
 	log.WithFields(log.Fields{
 		"ndiff": len(diffs),
 	}).Debugf("umoci: checked mtree spec")
+
+	// We need to mask config.Volumes.
+	config, err := mutator.Config(context.Background())
+	if err != nil {
+		return errors.Wrap(err, "get config")
+	}
+	maskedPaths := ctx.StringSlice("mask-path")
+	if !ctx.Bool("no-mask-volumes") {
+		for v := range config.Volumes {
+			maskedPaths = append(maskedPaths, v)
+		}
+	}
+	diffs = mtreefilter.FilterDeltas(diffs, mtreefilter.MaskFilter(maskedPaths))
 
 	reader, err := layer.GenerateLayer(fullRootfsPath, diffs, &meta.MapOptions)
 	if err != nil {
--- a/pkg/mtreefilter/mask.go
+++ b/pkg/mtreefilter/mask.go
@@ -1,0 +1,76 @@
+/*
+ * umoci: Umoci Modifies Open Containers' Images
+ * Copyright (C) 2017 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


cyphar:
@vrothberg PTAL

vrothberg:
Sorry, I did not forget about this. Hopefully I find time to check this PR next week.

cyphar:
This is now actually recommended by opencontainers/image-spec#694 (yay, finally).

vrothberg:
Awesome! I put checking the code on tomorrow's to-do.

cmd/umoci/repack.go
```diff
@@ -161,9 +172,11 @@ func repack(ctx *cli.Context) error {
...
 	}
-	var volumes []string
-	for v := range config.Volumes {
-		volumes = append(volumes, v)
+	volumes := ctx.StringSlice("mask-path")
+	if ctx.Bool("mask-volumes") {
+		for v := range config.Volumes {
+			volumes = append(volumes, v)
+		}
 	}
```
vrothberg: I suggest to rename `volumes` to `maskedPaths`.

test/repack.bats
```diff
@@ -499,3 +499,82 @@ function teardown() {
...
+	! [ -e "$BUNDLE_B/rootfs/volume/test" ]
+	! [ -e "$BUNDLE_B/rootfs/some nutty/path name" ]
+	! [ -e "$BUNDLE_B/rootfs/some nutty/path name/ here" ]
+
+	# Re-do everything but this time with --mask-path.
+	umoci repack --image "${IMAGE}:${TAG}-new" --mask-path /volume "$BUNDLE_A"
+	[ "$status" -eq 0 ]
+	image-verify "${IMAGE}"
+
+	# Re-extract to verify the volume changes weren't included.
```
vrothberg: ...the masked path changes...

test/repack.bats
```diff
@@ -499,3 +499,82 @@ function teardown() {
...
+	[ -d "$BUNDLE_C/rootfs/some/volume" ]
+	[ -f "$BUNDLE_C/rootfs/some/volume/test" ]
+	[[ "$(cat "$BUNDLE_C/rootfs/some/volume/test")" == "in a mirror directory" ]]
+	[ -f "$BUNDLE_C/rootfs/volumetest" ]
+	[[ "$(cat "$BUNDLE_C/rootfs/volumetest")" == "checking mirror" ]]
+
+	# Masked paths must not be included.
+	! [ -e "$BUNDLE_C/rootfs/volume" ]
+	! [ -e "$BUNDLE_C/rootfs/volume/test" ]
+	# But volumes might not be included.
```
vrothberg: might BE included?

pkg/mtreefilter/mask.go
```diff
@@ -0,0 +1,81 @@
...
+	"path/filepath"
+
+	"github.com/apex/log"
+	"github.com/vbatts/go-mtree"
+)
+
+// FilterFunc is a function used when filtering deltas with FilterDeltas.
+type FilterFunc func(path string) bool
+
+// isParent returns whether the path a is lexically an ancestor of the path b.
```
cyphar: Thanks. :wink:

pkg/mtreefilter/mask.go
```diff
@@ -0,0 +1,81 @@
...
+	return func(path string) bool {
+		// Convert path to relative-to-root.
+		path = filepath.Clean(path)
+		path = filepath.Join("/", path)
+
+		// Check that no masks are matched.
+		for _, mask := range masks {
+			// Mask also needs to be relative-to-root.
+			mask = filepath.Clean(mask)
+			mask = filepath.Join("/", mask)
```
cyphar: See above.

test/repack.bats
```diff
@@ -499,3 +499,56 @@ function teardown() {
...
+	[ "$status" -eq 0 ]
+	image-verify "${IMAGE}"
+
+	# Unpack the image.
+	umoci unpack --image "${IMAGE}:${TAG}" "$BUNDLE_A"
+	[ "$status" -eq 0 ]
+	bundle-verify "$BUNDLE_A"
+
+	# Create files in those volumes.
+	mkdir -p "$BUNDLE_A/rootfs/some nutty/path name/"
```
cyphar: Hmmm. I'm not sure. `VOLUME` in a Dockerfile creates the directory in the image (from memory) so I'm not sure if it makes sense for us to do that -- what should we do if a non-directory already exists there (for example)? I don't like the idea of erroring out in that case, but ignoring it also seems fishy.

I'll have to think about it some more, but can you open an issue for us to review at a later date?

test/repack.bats
```diff
@@ -499,3 +499,56 @@ function teardown() {
...
+	[ -f "$BUNDLE_B/rootfs/volumetest" ]
+	[[ "$(cat "$BUNDLE_B/rootfs/volumetest")" == "checking mirror" ]]
+
+	# Volume paths must not be included.
+	! [ -e "$BUNDLE_B/rootfs/volume" ]
+	! [ -e "$BUNDLE_B/rootfs/volume/test" ]
+	! [ -e "$BUNDLE_B/rootfs/some nutty/path name" ]
+	! [ -e "$BUNDLE_B/rootfs/some nutty/path name/ here" ]
+
+	# TODO: We need to add some flags to repack to allow users to ignore VOLUME.
```
vrothberg: There's one hiding at the bottom :)

test/repack.bats
vrothberg: Good idea, let's move to an issue :+1:

cyphar:
I'm going to wait on merging this until I speak to @davidcassany about how he feels about `--no-mask-volumes` and how we're going to handle CLI changes in the future.

cyphar:
I've rebased this and made it opt-out after discussions with @davidcassany. I agree with him that I severely doubt that any current users will be broken by this and in the future KIWI will have strict requirements on `umoci` versions.

cyphar:
LGTM.

Issue #126, opened 2017-06-02
*: switch to golang.org/x/sys/unix


cyphar:
They provide wrappers for most of the code in `system`. However I'm not sure how we'll have to handle `unpriv`. On the plus side we'll have less cross-`pkg` linking.

PR #128, opened 2017-06-15
*: switch to using golang.org/x/sys/unix

cyphar:
This is necessary for the repeal and replacement of pkg/system with
golang.org/x/sys. Believe me, it's going to be HUGE.

Implements #126 


cyphar:
Found a fun bug in Go's stdlib. golang/go#20698. So I'll just comment out the empty-value code.

cyphar:
The empty value tests are actually breaking **other** `xattr`s. It looks like there's some `go-mtree` bugs involved as well as some odd things going on with `archive/tar`.

PR #131, opened 2017-06-29
ci: fix EPERM coverage file

cyphar:
This recently broke on Travis, and it appears that there's some
mysterious permission issue happening with the interactions between
Docker and the coverage file.

Really we should make the CI not use Docker at all, but the package
requirements are forcing us to use CI for now. Hopefully we can improve
this in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ notifications:
     email: false
 
 script:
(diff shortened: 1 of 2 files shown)


cyphar:
Weird, the CI isn't being kicked off...

cyphar:
LGTM.

PR #133, opened 2017-06-30
*: update to image-spec v1.0.0-rc5

### LINKED ISSUE #132: tests: remove need for Docker
Usage of Docker is a bit of an anti-pattern IMO, because it actually makes it harder to run the tests in environments where you don't have Docker (such as inside `rpmbuild`). Luckily the bats tests aren't *too* specific to Docker.

* [ ] Make a nice way of getting the right version of our various testing dependencies.
* [ ] Set up a pulling script for images that operates outside of Docker.
* [ ] Clean up the weird mess of environment variable passing in the Makefile.
* [ ] (optional) Make testing more modular so that you can run all tests that don't require internet (or something).

The only downside to doing this is that the CI will no longer be running inside openSUSE so we can't make sure our packages are sane (the upside is that we no longer have to get packages into OBS in order to test with them on the CI).

Also Docker is the main reason that #131 is necessary.

cyphar:
This is a fairly large semantic and library change because of the semantic
change of tags no longer being a first-class property of the underlying
storage. This required a change to `oci/cas` and adding reference resolution to
`oci/casext` (which is actually nicer on some level but conflicts are hard to
resolve now).

Currently platform filtering is not implemented, and the current handling of
ambiguous tags is not really ironed out (errors are returned). This is
something that will need to be ironed out in the future especially if we want
to support creating multi-level trees of `Index`es.

In addition, due to the way that our CI works we need to update to the latest
`skopeo` but we can't do this in the distribution because this change is still
being worked on. IMO it would be better if we first resolve #132 before merging
this so that we actually can test this in a more sane way going forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   program][cii]. openSUSE/umoci#134
 - `umoci` also now has more extensive architecture, quick-start and roadmap
   documentation. openSUSE/umoci#134
+- `umoci` now supports [`1.0.0-rc5` of the OCI image
+  specification][ispec-v1.0.0-rc5]. Note that there are still some remaining UX
+  issues with `--image` and other parts of `umoci` which may be subject to
+  change in future versions. In particular, this update of the specification
+  now means that images may have ambiguous tags. `umoci` will warn you if an
+  operation may have an ambiguous result, but we plan to improve this
+  functionality far more in the future. openSUSE/umoci#133
+
 ### Changed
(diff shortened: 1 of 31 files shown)


cyphar:
I will follow up with a `1.0.0-rc6` bump immediately after this is merged, as the differences are far more minor.

cyphar:
LGTM.

cyphar:
LGTM. There's plenty of improvements left that are necessary but this works for the moment.

cyphar:
I just realised that the tests passed even though `image-tools` hasn't been updated in openSUSE. That's pretty worrying, and I'll update `image-tools` for #142.

PR #140, opened 2017-07-11
make: add validation for reproducibility

cyphar:
Add a new target local-validate-reproducible which will verify that our
builds remain reproducible even if you rebuild them within 10s of each
other and the modified time of the source code is changed. This is a
very trivial way of checking reproducibility, but Go makes guarantees
that the binaries are always reproducible[1].

[1]: https://blog.filippo.io/reproducing-go-binaries-byte-by-byte/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,4 +28,5 @@ script:
   - chmod a+rwx . # Necessary to make Travis co-operate with Docker.
   - make umoci
   - make umoci.static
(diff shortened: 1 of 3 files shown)


cyphar:
This is necessary for silver CII badging.

cyphar:
LGTM

cyphar:
LGTM

PR #142, opened 2017-07-15
OCI final RC bump

cyphar:
This is the follow-up to #133 and includes an update to the final `-rc` versions for both `image-spec` and `runtime-spec`. I will do an immediate follow-up once OCI 1.0.0 is released next week.


cyphar:
LGTM.

PR #145, opened 2017-07-16
mutate: rework to properly handle arbitrary paths

cyphar:
This involves creating the concept of a `DescriptorPath` which will allow for 
`mutate` to simply recreate all of the blobs (in reverse order).

Currently WIP.


cyphar:
Currently the `casext.MapDescriptors` doesn't work. Something odd is happening when you have non-slices. I think I'll have to start passing around pointers to fields or similar.

cyphar:
LGTM.

PR #149, opened 2017-07-19
*: update to v1.0.0 of {runtime,image}-spec

cyphar:
On paper this change should not affect anything, but we needed to add a
patch in order to fix a late change in runtime-spec that broke the
runtime-tools/generate build.


cyphar:
LGTM.

cyphar:
LGTM. We will have to handle the `runtime-tools` update properly later.

PR #158, opened 2017-07-23
	*: rework validation scripts

cyphar:
It appears that local-validate has not been running in recent PRs. Fix
this by reworking `make ci` and the travis configuration to make sure it
runs. This is critical for CII bading.

In addition, this un-comments git-validation which has resolved their
issues with running in Travis.


cyphar:
LGTM.

PR #159, opened 2017-07-24
doc: add security considerations

cyphar:
This adds a document that describes the current security considerations
and trade-offs made within umoci. It is not a complete document (and two
of the three issues have yet to be satisfactorily addressed) but it
helps users understand what the current situation is (as well as
providing some information to potential contributors).


cyphar:
LGTM.

PR #162, opened 2017-07-27
*: fix xattr inconsistency issue

### LINKED ISSUE #161: xattr: empty xattrs appear to break unrelated tar entries
When adding some extra tests in #141, I noticed that some cases will actually cause `go-mtree` to ignore some `xattr`s. In particular, adding some empty-valued `xattr`s would cause _unrelated_ extended attributes to not be included.

It's possible that this was fixed by vbatts/go-mtree#135, but we need to double-check.

cyphar:
Fixes #161


cyphar:
LGTM.

PR #163, opened 2017-08-11
release: make script more generic

cyphar:
This is so that other projects (such as runc) can use our release script
with minimal changes (and the changes can be contained to just the top
portion of the script, making updating easier).


cyphar:
LGTM.

PR #167, opened 2017-08-25
repack: change --[ug]id-map= UX to match user_namespaces(7)

cyphar:
Our IDMap interfaces were not matching what a user who is familiar with
user_namespaces(7) would expect. This caused me some confusion while
trying to debug a user namespace bug. While this is a breaking change,
I'm fairly confident that there are almost no users of this feature.

In addition, actually add documentation on the --[ug]id-map format
(which was previously not explicitly documented at all).

Fixes: 9f88dd9 ("{un,re}pack: implement ID mapping")


cyphar:
LGTM.

PR #170, opened 2017-09-09
docker build fails

ryanolson:
simple fix to the dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,6 @@ RUN zypper -n in \
 		git \
 		'go>=1.6' \
 		golang-github-cpuguy83-go-md2man \
-		go-mtree \
 		jq \
 		make \
 		moreutils \


cyphar:
Thanks, but we need `go-mtree` for the tests. I guess I must've dropped it from an openSUSE repo without noticing. I'll fix that up.

ryanolson:
go-mtree is in the vendor directory, so you include it, just not via the package manager.

cyphar:
@ryanolson We also use it as a binary, not just as a library. [In the tests, we do verification that `gomtree` succeeds after unpacking and so on.](https://github.com/openSUSE/umoci/blob/599ec5b416b25fce508249bd86e7bb62cf1e3308/test/unpack.bats#L50)

cyphar:
This has been fixed in #176. Thanks for your patience, there were a lot of issues to deal with.

PR #176, opened 2017-09-21
ci: update Dockerfile to fix broken tests

cyphar:
Due to some weird py3k issue that has manifested recently, our CI
started breaking due to errors within xattr. Switch to a newer version
(since it looks like the python3-* version is not maintained anymore in
the py3k repo).


cyphar:
`xattr` breakage is upstream (tracked in vbatts/go-mtree#140).

cyphar:
LGTM.

cyphar:
This also includes a backport of opencontainers/runtime-tools#491, which is necessary because the current validator violates the spec.

PR #177, opened 2017-09-29
oci: remove 'drivers' abstraction

cyphar:
This effectively soft-reverts 184ab749f39a ("oci: cas: switch to
driver-based model for CAS"), which was effectively a mistake. The
concept of drivers doesn't make too much sense -- we are unlikely to
ever have more than one driver (unless we add remote backends).


cyphar:
LGTM.

cyphar:
(@hallyn: This is one of the changes that I mentioned.)

hallyn:
Seems reasonable.

cyphar:
LGTM.

PR #178, opened 2017-09-29
oci: consume any excess data present in layer archives

cyphar:
Different tar implementations can have different levels of redundant
padding and other similar weird behaviours. While on paper they are all
entirely valid archives, Go's tar.Reader implementation doesn't
guarantee that the entire stream will be consumed (which can result in
the later diff_id check failing because the digester didn't get the
whole uncompressed stream). Just blindly consume anything left in the
layer.

Reported-by: "Serge E. Hallyn" <serge@hallyn.com>


hallyn:
Thanks - tested for awhile now and works great.

Tested-by: Serge Hallyn <serge@hallyn.com>

cyphar:
LGTM. I'll write a test up for this usecase once we either have `umoci raw repack-layer` or something like that.

PR #182, opened 2017-10-03
layer: always set bundle to 0700 on creation

### LINKED ISSUE #181: [debatable] Make rootfs 700?
This could fall under "don't shoot yourself in the foot", but if I

```bash
umoci unpack  --image oci:alpine alpine
```

then alpine/rootfs is world descendable.  This means if there are any setuid-root exploitable binaries in the roots, an unprivileged user may be able to get to them.

Like I say, I understand if you choose to ignore it, but I thought I should point it out.

cyphar:
This protects against attacks where an unsafe image is extracted into a
location where an unprivileged user could potentially access unsafe
setuid binaries (or unsafe inodes). This protection is very easy to
disable by a user of umoci, but the intention is to protect against
lowest-common-denominator attacks.

This is a slightly breaking change, but as it is a security issue (which
was previously documented) we have made the change anyway.

Fixes #181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   switch to the trivially different (but now more consistent) format.
   openSUSE/umoci#167
 
+### Security
+- `umoci unpack` used to create the bundle and rootfs with world
+  read-and-execute permissions by default. This could potentially result in an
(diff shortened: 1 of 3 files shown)


cyphar:
I really should've done this earlier. I was worried about breaking the usecase of a shared rootfs, which in retrospect wasn't a good idea.

/cc @hallyn

cyphar:
LGTM.

PR #187, opened 2017-10-05
unpack: cleanup bundle in case of failure

### LINKED ISSUE #174: clean up after failed unpack
If you 'umoci unpack --image x:y dest', and there is an error in the x:y config, then dest will remain on disk, partially unpacked.  There may be reasons not to, but offhand it seems most intuitive to clean up.

cyphar:
In order to ensure we don't leave around a broken bundle directory,
remove the bundle if a failure occurred. While cleaning up isn't a huge
issue for unpacking as root, cleaning up as an unprivileged user gets
quite complicated ("rm -rf" doesn't work).

Fixes #174 
Reported-by: "Serge E. Hallyn" <serge@hallyn.com>

/cc @hallyn


cyphar:
LGTM.

PR #190, opened 2017-10-11
*: add ability to use a shared CAS directory

jonboulle:
umoci currently always expects blobs to live within the image layout
it's working on, in the aptly-named "blobs" subdirectory. However, the
OCI specification [1][1] permits this directory to be empty and for
implementations to look in other locations for referenced blobs.
This is particularly useful when working with multiple image layouts and
wanting to share blobs between them.

This patch adds a `--shared-cas` flag to all umoci commands that work
with image layouts. The flag expects a directory to be passed, and if
set, this directory will be used for all blob operations.

In the first implementation, we extend the `dirEngine` implementation of
the CAS interface to accept another (optional) `sharedCasPath` field.
Analogously to the CLI flag, if this field is supplied, it will be used
internally by the dirEngine for all blob-related operations.

[1]: https://github.com/opencontainers/image-spec/blob/7c889fafd04a893f5c5f50b7ab9963d5d64e5242/image-layout.md#blobs

### DIFF
--- a/oci/cas/cas.go
+++ b/oci/cas/cas.go
@@ -32,13 +32,18 @@ import (
 )
 
 const (
-	// BlobAlgorithm is the name of the only supported digest algorithm for blobs.
-	// FIXME: We can make this a list.
-	BlobAlgorithm = digest.SHA256
+	// DefaultBlobAlgorithm is the default supported digest algorithm for blobs
+	DefaultBlobAlgorithm = digest.SHA256
 )
 
-// Exposed errors.
 var (
+	// BlobAlgorithms contains the supported digest algorithms for blobs
+	BlobAlgorithms = []digest.Algorithm{
+		digest.SHA256,
+	}
+
+	// Exposed errors
+
 	// ErrNotExist is effectively an implementation-neutral version of
 	// os.ErrNotExist.
 	ErrNotExist = fmt.Errorf("no such blob or index")
--- a/cmd/umoci/config.go
+++ b/cmd/umoci/config.go
@@ -123,6 +123,7 @@ func parseKV(input string) (string, string, error) {
 func config(ctx *cli.Context) error {
 	imagePath := ctx.App.Metadata["--image-path"].(string)
 	fromName := ctx.App.Metadata["--image-tag"].(string)
+	sharedCasPath := ctx.String("shared-cas")
 
 	// By default we clobber the old tag.
 	tagName := fromName
@@ -131,7 +132,7 @@ func config(ctx *cli.Context) error {
 	}
 
 	// Get a reference to the CAS.
-	engine, err := dir.Open(imagePath)
+	engine, err := dir.Open(imagePath, sharedCasPath)
 	if err != nil {
 		return errors.Wrap(err, "open CAS")
 	}
--- a/cmd/umoci/gc.go
+++ b/cmd/umoci/gc.go
@@ -51,9 +51,10 @@ root set of references. All other blobs will be removed.`,
 
 func gc(ctx *cli.Context) error {
 	imagePath := ctx.App.Metadata["--image-path"].(string)
+	sharedCasPath := ctx.String("shared-cas")
 
 	// Get a reference to the CAS.
-	engine, err := dir.Open(imagePath)
+	engine, err := dir.Open(imagePath, sharedCasPath)
 	if err != nil {
 		return errors.Wrap(err, "open CAS")
 	}
--- a/cmd/umoci/gc.go
+++ b/cmd/umoci/gc.go
@@ -51,10 +51,12 @@ root set of references. All other blobs will be removed.`,
 
 func gc(ctx *cli.Context) error {
 	imagePath := ctx.App.Metadata["--image-path"].(string)
-	sharedCasPath := ctx.String("shared-cas")
+	if ctx.String("shared-cas") != "" {
+		return errors.New("gc not supported with --shared-cas")
+	}
 
 	// Get a reference to the CAS.
-	engine, err := dir.Open(imagePath, sharedCasPath)
+	engine, err := dir.Open(imagePath, "")
 	if err != nil {
 		return errors.Wrap(err, "open CAS")
 	}
--- a/cmd/umoci/new.go
+++ b/cmd/umoci/new.go
@@ -54,9 +54,10 @@ needing a base image to start from.`,
 func newImage(ctx *cli.Context) error {
 	imagePath := ctx.App.Metadata["--image-path"].(string)
 	tagName := ctx.App.Metadata["--image-tag"].(string)
+	sharedCasPath := ctx.String("shared-cas")
 
 	// Get a reference to the CAS.
-	engine, err := dir.Open(imagePath)
+	engine, err := dir.Open(imagePath, sharedCasPath)
 	if err != nil {
 		return errors.Wrap(err, "open CAS")
 	}
--- a/cmd/umoci/raw-runtime-config.go
+++ b/cmd/umoci/raw-runtime-config.go
@@ -87,6 +87,7 @@ func rawConfig(ctx *cli.Context) error {
 	imagePath := ctx.App.Metadata["--image-path"].(string)
 	fromName := ctx.App.Metadata["--image-tag"].(string)
 	configPath := ctx.App.Metadata["config"].(string)
+	sharedCasPath := ctx.String("shared-cas")
 
 	var meta UmociMeta
 	meta.Version = UmociMetaVersion
@@ -124,7 +125,7 @@ func rawConfig(ctx *cli.Context) error {
 	}).Debugf("parsed mappings")
 
 	// Get a reference to the CAS.
-	engine, err := dir.Open(imagePath)
+	engine, err := dir.Open(imagePath, sharedCasPath)
 	if err != nil {
 		return errors.Wrap(err, "open CAS")
 	}
--- a/cmd/umoci/repack.go
+++ b/cmd/umoci/repack.go
@@ -99,6 +99,7 @@ func repack(ctx *cli.Context) error {
 	imagePath := ctx.App.Metadata["--image-path"].(string)
 	tagName := ctx.App.Metadata["--image-tag"].(string)
 	bundlePath := ctx.App.Metadata["bundle"].(string)
+	sharedCasPath := ctx.String("shared-cas")
 
 	// Read the metadata first.
 	meta, err := ReadBundleMeta(bundlePath)
@@ -117,7 +118,7 @@ func repack(ctx *cli.Context) error {
 	}
 
 	// Get a reference to the CAS.
-	engine, err := dir.Open(imagePath)
+	engine, err := dir.Open(imagePath, sharedCasPath)
 	if err != nil {
 		return errors.Wrap(err, "open CAS")
 	}
--- a/cmd/umoci/stat.go
+++ b/cmd/umoci/stat.go
@@ -58,9 +58,10 @@ humans to read, and might change in future versions.`,
 func stat(ctx *cli.Context) error {
 	imagePath := ctx.App.Metadata["--image-path"].(string)
 	tagName := ctx.App.Metadata["--image-tag"].(string)
+	sharedCasPath := ctx.String("shared-cas")
 
 	// Get a reference to the CAS.
-	engine, err := dir.Open(imagePath)
+	engine, err := dir.Open(imagePath, sharedCasPath)
 	if err != nil {
 		return errors.Wrap(err, "open CAS")
 	}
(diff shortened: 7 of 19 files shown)


jonboulle:
@cyphar this is a bit of a POC or strawman, would love some feedback on whether you'd take a feature like this and how you'd expect to see it implemented (I can think of about a thousand different approaches so not sure which you'd prefer).

wking:
On Wed, Oct 11, 2017 at 02:53:52PM +0000, Jonathan Boulle wrote:
> This is particularly useful when working with multiple image layouts
> and wanting to share blobs between them.

I'm not sure if umoci is interested in picking this up, but as a
related idea has a CAS-protocol registry [1] with an
oci-cas-template-v1 entry [2].  The URI Template approach is inspired
by parcel (e.g. [3]).  With a similar URI Template approach for refs
[4], image-spec layouts can become a special-case ref-engine-discovery
approach (xiekeyang/oci-discovery#20).  With that approach, you could
add casEngines entries to the oci-layout file [5] to point at your
central CAS engine:

  $ cat oci-layout
  {
    "imageLayoutVersion": "1.1.0",
    "refEngines": [
      {
        "protocol": "oci-index-template-v1",
        "uri": "index.json"
      }
    ],
    "casEngines": [
      {
        "protocol": "oci-cas-template-v1",
        "uri": "file:///wherever/you/keep/your/blobs/{algorithm}/{encoded}"
      }
    ]
  }

> This patch adds a `--shared-cas` flag to all umoci commands that
> work with image layouts.

The casEngines-in-oci-layout approach lets you declare this sort of
thing in the layout itself, where all consumers can access it.  That
means you'd only need a --shared-cas entry when you initially create a
new layout.

If there is any umoci interest in this approach, I'm happy to help
code it up.  oci-index-template-v1 is already implemented in Go [6],
and I was planning on writing an oci-cas-template-v1 implementation
this week anyway (I just hadn't decided on which repo to put it in yet
;).

[1]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/cas-engine-protocols.md
[2]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/cas-template.md
[3]: https://github.com/cyphar/parcel/commit/b10625235275423b27032b764d8888e76aa2b122#diff-6a735bf86dcc8fb712e11c207df01ae4R88
[4]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/index-template.md
[5]: https://github.com/wking/image-spec/blob/ref-engine-discovery-layout/image-layout.md#layout-object
[6]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/tools/refengine/indextemplate/indextemplate.go

cyphar:
@wking As I've mentioned before, my issue with `casEngines` is because of precisely the layer violation you are referring to here. I've rewritten significant portions of parcel's specification, by the way.

wking:
On Thu, Oct 12, 2017 at 07:41:23AM +0000, Aleksa Sarai wrote:
> @wking As I've mentioned before, my issue with `casEngines` is
> because of precisely the layer violation you are referring to here.

I expect you're referencing [1].  Can you provide more details on the
issue you see (possibly in an oci-discovery issue, if you feel it's
too off-topic here)?  Is it the lack of blob-media-type routing you
added to parcel [2]?  I agree that the OCI CAS Template Protocol [3]
does not address that, but there's no reason you couldn't register [4]
a new CAS-engine protocol that addresses it.  For example, if we
registered your parcel blobURIs as parcel-blobURIs-v1, you could have
a casEngines entry like:

  "casEngines": [
    {
      "protocol": "parcel-blobURIs-v1",
      "blobURIs": [
        {
          "mediaType": "application/vnd.oci.image.manifest.v1+json",
          "templates": [
            "https://manifests.cyphar.com/opensuse/{parcel.discovery.nameDigest}/manifests/{parcel.fetch.blob.digestAlgorithm}/{parcel.fetch.blob.digest}"
          ]
        },
        …
      ]
    },
    …
  ]

And for things like Docker's registry and its auth dance, you could
have a ‘docker’ CAS-engine protocol similar to the ‘docker’ ref-engine
protocol I sketched in [5].

> I've rewritten significant portions of parcel's specification, by
> the way.

Glad to see that moving forward :).  I filed some small copy-edit PRs
a few hours back ;).

[1]: https://github.com/xiekeyang/oci-discovery/issues/1#issuecomment-327559804
[2]: https://github.com/cyphar/parcel/blame/b10625235275423b27032b764d8888e76aa2b122/DESIGN.md#L335-L339
[3]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/cas-template.md
[4]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/cas-engine-protocols.md
[5]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/well-known-uri-ref-engine-discovery.md#example-1

cyphar:
@wking 

> I expect you're referencing [1].  Can you provide more details on the
issue you see (possibly in an oci-discovery issue, if you feel it's
too off-topic here)?

My issue is that it doesn't "smell" right for semantic information about this sort of "shared CAS" to be stored in descriptors. I'm a huge proponent of doing it top-down, distribution style. Effectively it boils down to that I'm not convinced that mis-using descriptors as a way of trying to do something more peer-to-peer is better in the long run than having a federated distribution system (similar to the existing distribution system for packages).

> Is it the lack of blob-media-type routing you
added to parcel [2]?

No, it's the above, but you're right that the reason I added multiple backend support was with the intention of allowing combining blobs at a higher level (similar to what the PR does, but at the pull stage and not at the actual operational stage).

But all of this is quite off-topic. If you want we can discuss this in the parcel issue tracker (or oci-discovery), or email.

cyphar:
@jonboulle I'm not opposed to the idea (as we discussed recently), but I'll have to get back to you on which layer this should be done. Also, out of interest, did you find any additional issues with symlinking blobs other than `umoci gc` -- or is this purely a make-users-lives-easier thing?

jonboulle:
@cyphar manual symlink munging works, but we definitely want to get out of that business if at all possible and punt whatever we can into the OCI tooling - especially if there's any chance that something _like_ opencontainers/image-tool#40 might happen. (You might consider this my attempt to create precedent, with a clear use case and active implementation..)

I should mention that since our particular use doesn't involve particularly long-running systems (i.e. there's a certain level of ephemerality in the CAS usage) we're happy for the implementation to change across some releases if we need to iterate towards a converged solution

oci/cas/cas.go
```diff
@@ -32,13 +32,18 @@ import (
 )
 
 const (
-	// BlobAlgorithm is the name of the only supported digest algorithm for blobs.
-	// FIXME: We can make this a list.
-	BlobAlgorithm = digest.SHA256
+	// DefaultBlobAlgorithm is the default supported digest algorithm for blobs
+	DefaultBlobAlgorithm = digest.SHA256
```
wking: Deprecate in favor of go-digest's [`Canonical`][1]?

[1]: https://github.com/opencontainers/go-digest/blob/v1.0.0-rc1/algorithm.go#L39

wking: > Deprecate in favor of go-digest's `Canonical`?

Spun off into #197.

wking:
> With that approach, you could add casEngines entries to the oci-layout file \[5] to point at your central CAS engine:

I've walked a bit down this path with wking/umoci@2908a4a5e78708e76cd01444b027802af884bc9c.  While I [still][1] think decoupling the CAS and ref engines is useful, I no longer think you can configure a digest-listing CAS engine with just a URI Template.  [URI Templates are not reversible in general][2] (jtacoma/uritemplates#2), and even if they were, there are issues with reversing [reference resolution][3].  You need  to reverse expansion to list stored digests ([casengines][4], [umoci][5]).  In casengine, I'm getting around that by requiring the caller to provide their own reverser (wking/casengine@76d40621518457fe0471f9689bcae3807cf6558a).  But since you don't need digest-listing support for most operations and because the base URI for relative expansion is unclear, including a reversing regexp pattern in [the engine config][6] seems like too large a hack.  That means any CAS engines [configured solely via `casEngines` in `oci-layout`][7] would not be able to list digests.

However, I'm still not sure we need to go as far as bdd8bdf44 and pass down location information from every command that might interact with CAS.  Operations like `umoci unpack …` which only need to pull blobs would be fine loading a template URI from `oci-layout`.  I'll continue working in this direction, passing the reversing regexp down from the callers in cases that need reversing logic, although I'm obviously just providing suggestions for umoci and not in charge of what gets landed.

If loading the CAS URI Template or base directory from `oci-layout` is off the table, an alternative to mitigate the tedium of `--shared-cas` entries would be to set [`EnvVar`][8] when declaring `--shared-cas`.  That would be convenient for umoci users, but unless the environment variable catches on it wouldn't help other tools trying to pull from layers that used an alternate blob location.

And while the `dirEngine.sharedCasPath` base-path approach is certainly simpler than using URI Templates, the base-path approach does not allow sharding (opencontainers/image-spec#449), while [the URI Template approach does][9].  I'm not sure if modern filesystems care all that much about sharding for directory listings, so maybe this particular point is not a strong argument either way.

[1]: https://github.com/openSUSE/umoci/commit/667271b57db5637a18b0f249026a84608f834b3d#commitcomment-19711538
[2]: https://tools.ietf.org/html/rfc6570#section-1.4
[3]: https://golang.org/pkg/net/url/#URL.ResolveReference
[4]: https://github.com/wking/casengine/blob/a94b57c99b5b2f85b5a256a589acf8d062461618/dir/dir.go#L153
[5]: https://github.com/openSUSE/umoci/blob/v0.3.1/oci/cas/dir/dir.go#L286
[6]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/cas-template.md
[7]: https://github.com/openSUSE/umoci/pull/190#issuecomment-335882368
[8]: https://github.com/urfave/cli/tree/v1.20.0#values-from-the-environment
[9]: https://github.com/xiekeyang/oci-discovery/blob/0be7eae246ae9a975a76ca209c045043f0793572/cas-template.md#example

cyphar:
@wking 

> URI Templates are not reversible in general (jtacoma/uritemplates#2), and even if they were, there are issues with reversing reference resolution. You need to reverse expansion to list stored digests (casengines, umoci).

I personally think that `ListBlobs` should be something that is optionally supported. In `parcel`, I envision that you might want to have some repos that have a list of the blobs somewhere (which could be given as a URI template in some extension) but that is not necessary for normal operations. We could make `umoci gc` no-op if `ListBlobs` gives nothing (which is not the end of the world).

Aside from `ListBlobs`, I still feel that a URI Template approach is much more flexible and allows you to specify stronger top-down policy decisions on a domain. However, `parcel` discussions are a bit out-of-place here. :wink:

wking:
On Thu, Oct 19, 2017 at 08:36:43AM -0700, Aleksa Sarai wrote:
> I personally think that `ListBlobs` should be something that is
> optionally supported.

Yeah.  This is where I was going with my second paragraph in [1].
I'll post again once I have a working implementation.

> However, `parcel` discussions are a bit out-of-place here. :wink:

This isn't really a parcel discussion.  I think a URI Template
approach to blob location would be a good way to address @jonboulle's
shared-CAS usecase *in umoci*.  wking/umoci@2908a4a, which I mentioned
in [1], is initial work in that direction, and I'll keep working on it
until I have a working example of the “loading a template URI from
oci-layout” idea for you to consider.  So I think these alternative
approaches are on-topic for this PR, although I'll try to keep my
references short.

Alternatively, I could open a WIP PR with my wking/umoci@2908a4a and
compare that approach with this one in that PR.  That divides the
conversation, but would keep URI Template discussion out of this
branch of the conversation.  Let me know if you want me to file that
alternative PR (which would not imply any approval of that PR's
direction).

[1]: https://github.com/openSUSE/umoci/pull/190#issuecomment-337742882

cyphar:
Sorry for the delays. I just re-read the spec and you're completely right that this is supported (though I anticipate that the reason for this wording is to allow for the "external artifact" feature that ACI has). I also now see why @wking was talking so much about `parcel` and `oci-discovery` -- it's quite clear if you stare at the text for a while that remote blob fetching would be a valid application (though I think for distribution `parcel` solves the problem better).

I will re-review this pull request next week (please note that I have some assignments and exams coming up, so I'll have less time to work than usual, so apologies upfront for any delays).

cyphar:
This is fairly stale by now, closing. I'd be happy to re-discuss this issue again if it's still an issue you're having with `runc`.

PR #191, opened 2017-10-12
unpack: detect nonexistent tag [trivial]

hallyn:
previously, nonexistent tags were confusingly reported as being
ambiguous.

Signed-off-by: Serge Hallyn <serge@hallyn.com>


cyphar:
LGTM, yeah oops.

PR #192, opened 2017-10-13
detect nonexistent tag

hallyn:
Previously, nonexistent tags were confusingly reported as being
ambiguous.

Commit 09c6ab38 only handled one instance, catch the rest.

Signed-off-by: Serge Hallyn <serge@hallyn.com>

### DIFF
--- a/cmd/umoci/config.go
+++ b/cmd/umoci/config.go
@@ -142,6 +142,9 @@ func config(ctx *cli.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "get descriptor")
 	}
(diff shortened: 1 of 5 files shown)


hallyn:
Hm, failed.  I'll have to look into that...

cyphar:
```
# runtime: address space conflict: map(0xc420100000) = 0x7f7311cb0000
# fatal error: runtime: address space conflict
```

Looks like some weird Go runtime bug to me. Restarting the test.

cyphar:
LGTM.

cyphar:
LGTM.

PR #194, opened 2017-10-17
pkg: system: use Major, Minor and Mkdev from x/sys/unix

tklauser:
As suggested by @cyphar in opencontainers/runc#1614 :wink: 

- re-vendor golang.org/x/sys to get Major, Minor and Mkdev
- use these functions in pkg/system


cyphar:
:smile_cat: Thanks!

LGTM.

PR #195, opened 2017-10-18
pkg: system: use UtimesNanoAt from x/sys/unix

tklauser:
Use UtimesNanoAt from golang.org/x/sys/unix instead of manually crafting
the syscall.

Signed-off-by: Tobias Klauser <tklauser@distanz.ch>


cyphar:
LGTM. Thanks. :heart:

PR #201, opened 2017-10-19
umoci/repack: add `--refresh-bundle` flag

### LINKED ISSUE #196: umoci repack --refresh-bundle
One usecase that @jonboulle had was that they want to avoid having to do something like:

```
% umoci repack --image blah:tag bundle_1
% umoci unpack --image blah:tag bundle_2
```

This could be combined into an operation that updates the state of `bundle` as though `umoci unpack` created said bundle. This just comes down to updating the `mtree` manifest and the descriptor path in `umoci.json`. The workflow would then just be:

```
% umoci repack --image blah:tag --refresh-bundle bundle_1
```

jonboulle:
When this flag is supplied, umoci will refresh the bundle's metadata
(namely, the mtree manifest and the umoci metadata JSON) to record the
state of the bundle when the repack is performed. This means that an
unpack immediately after a repack should be essentially idempotent.

Fixes #196

### DIFF
--- a/cmd/umoci/repack.go
+++ b/cmd/umoci/repack.go
@@ -130,7 +130,7 @@ func repack(ctx *cli.Context) error {
 		return errors.Wrap(err, "create mutator for base image")
 	}
 
-	mtreeName := strings.Replace(meta.From.Descriptor().Digest.String(), "sha256:", "sha256_", 1)
+	mtreeName := strings.Replace(meta.From.Descriptor().Digest.String(), ":", "_", 1)
 	mtreePath := filepath.Join(bundlePath, mtreeName+".mtree")
 	fullRootfsPath := filepath.Join(bundlePath, layer.RootfsName)
 
@@ -242,7 +242,7 @@ func repack(ctx *cli.Context) error {
 	log.Infof("created new tag for image manifest: %s", tagName)
 
 	if ctx.Bool("refresh-bundle") {
-		newMtreeName := strings.Replace(newDescriptorPath.Descriptor().Digest.String(), "sha256:", "sha256_", 1)
+		newMtreeName := strings.Replace(newDescriptorPath.Descriptor().Digest.String(), ":", "_", 1)
 		if err := generateBundleManifest(newMtreeName, bundlePath, fsEval); err != nil {
 			return errors.Wrap(err, "write mtree metadata")
 		}
--- a/cmd/umoci/repack.go
+++ b/cmd/umoci/repack.go
@@ -243,7 +243,7 @@ func repack(ctx *cli.Context) error {
 
 	if ctx.Bool("refresh-bundle") {
 		newMtreeName := strings.Replace(newDescriptorPath.Descriptor().Digest.String(), "sha256:", "sha256_", 1)
-		if err := writeMtree(newMtreeName, bundlePath, fsEval); err != nil {
+		if err := generateBundleManifest(newMtreeName, bundlePath, fsEval); err != nil {
 			return errors.Wrap(err, "write mtree metadata")
 		}
 		if err := os.Remove(mtreePath); err != nil {
--- a/cmd/umoci/repack.go
+++ b/cmd/umoci/repack.go
@@ -75,6 +75,10 @@ manifest and configuration information uses the new diff atop the old manifest.`
 			Name:  "no-mask-volumes",
 			Usage: "do not add the Config.Volumes of the image to the set of masked paths",
 		},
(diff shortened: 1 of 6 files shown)


cmd/umoci/utils.go
```diff
@@ -245,3 +246,37 @@ func Stat(ctx context.Context, engine casext.Engine, manifestDescriptor ispec.De
 
 	return stat, nil
 }
+
+// writeMtree writes an mtree of the rootfs in the given bundle path into the
+// file by the given name, using the supplied fsEval method
+func writeMtree(mtreeName string, bundlePath string, fsEval mtree.FsEval) error {
```
cyphar: I'd prefer this be named `generateBundleManifest` or something like that? Also if we're going to do the `: -> _` rename we might as well do it inside this function.

cmd/umoci/repack.go
```diff
@@ -75,6 +75,10 @@ manifest and configuration information uses the new diff atop the old manifest.`
 			Name:  "no-mask-volumes",
 			Usage: "do not add the Config.Volumes of the image to the set of masked paths",
 		},
+		cli.BoolFlag{
+			Name:  "refresh-bundle",
+			Usage: "update the bundle metadata (mtree) to reflect the packed rootfs",
```
cyphar: Please update the `man` pages. Also, you can drop the `(mtree)` part.

jonboulle:
> 4. gomtree $BUNDLE_A/rootfs to verify there's no changes in the mtree.

When you say "no changes", you mean between the new mtree and the bundle rootfs..?

cyphar:
Yeah, so it would be something like:

```
gomtree -p $BUNDLE_A/rootfs -f $BUNDLE_A/*.mtree
[ "$status" -eq 0 ]
[ "${#lines[@]}" -eq 0]
```

The idea being to ensure that the `mtree` has been correctly updated and that if we created a new layer it would be empty (to avoid duplication).

cyphar:
Oh, and this needs a `CHANGELOG` entry.

cmd/umoci/repack.go
```diff
@@ -236,5 +240,20 @@ func repack(ctx *cli.Context) error {
...
+	if ctx.Bool("refresh-bundle") {
+		newMtreeName := strings.Replace(newDescriptorPath.Descriptor().Digest.String(), "sha256:", "sha256_", 1)
+		if err := generateBundleManifest(newMtreeName, bundlePath, fsEval); err != nil {
+			return errors.Wrap(err, "write mtree metadata")
+		}
+		if err := os.Remove(mtreePath); err != nil {
+			return errors.Wrap(err, "remove old mtree metadata")
+		}
+		meta.From = newDescriptorPath
+		if err := WriteBundleMeta(bundlePath, meta); err != nil {
```
wking: I'm not sure how mtree storage works yet, so please ignore if this is a false alarm, but I think we want:

1. Generate new mtree.
2. Write bundle metadata pointing at new mtree.
3. Remove old mtree.

That way,we always have a working state, because dying with the old value unremoved seems better than dying with metadata pointing at a removed old value.

cmd/umoci/repack.go
```diff
@@ -236,5 +240,20 @@ func repack(ctx *cli.Context) error {
 	}
 
 	log.Infof("created new tag for image manifest: %s", tagName)
+
+	if ctx.Bool("refresh-bundle") {
+		newMtreeName := strings.Replace(newDescriptorPath.Descriptor().Digest.String(), "sha256:", "sha256_", 1)
```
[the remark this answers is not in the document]
jonboulle: fair point, done

cmd/umoci/repack.go
jonboulle: That's exactly what's happening here; `generateBundleManifest` is doing your 1 and 2. `WriteBundleMeta` is different metadata. Blame Aleksa for the names ;-)

doc/man/umoci-repack.1.md
```diff
@@ -62,6 +62,11 @@ The global options are defined in **umoci**(1).
   the image. This must be an ISO8601 formatted timestamp (see **date**(1)). If
   unspecified, the current time is used.
 
+**--refresh-bundle**
```
cyphar: You'll need to add it to the `SYNOPSIS` section as well.

jonboulle:
This should be good to go, but travis seems to be stuck somewhere?

cyphar:
LGTM. Thanks a lot @jonboulle! :smile_cat:

PR #203, opened 2017-10-23
Makefile: remove volume mount of source directory

jonboulle:
_[update]_

With the current Makefile, the Docker variants of the tests don't work
out of the box on SELinux-enabled systems because the volume mount
(namely, the source directory) isn't labelled correctly.

However we're already copying in the source, and the volume mount (over
the top of it) was only really used for `make shell`. Since this is an
uncommon use case, let's just drop the volume mount completely.

--

_[original PR, now at #206]_
With the current Makefile, the Docker variants of the tests don't work
out of the box on SELinux-enabled systems because the volume mount
(namely, the source directory) isn't labelled correctly. This patch
simply disables SELinux labelling for these particular containers,
as the easiest way to mitigate this issue. (This seems preferable to
using the `:z` or `:Z` option to the Docker volume mount, which would
forceably relabel the directory every time the Docker command is run.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,6 +87,5 @@ ENV SOURCE_IMAGE=/opensuse SOURCE_TAG=latest
 ARG DOCKER_IMAGE=opensuse/amd64:tumbleweed
 RUN skopeo copy docker://$DOCKER_IMAGE oci:$SOURCE_IMAGE:$SOURCE_TAG
 
-VOLUME ["/go/src/github.com/openSUSE/umoci"]
 WORKDIR /go/src/github.com/openSUSE/umoci
 COPY . /go/src/github.com/openSUSE/umoci
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ CMD := ${PROJECT}/cmd/umoci
 
 # We use Docker because Go is just horrific to deal with.
 UMOCI_IMAGE := umoci_dev
+DOCKER_RUN := docker run --rm -it
 
 # Output directory.
 BUILD_DIR ?= .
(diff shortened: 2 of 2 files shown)


jonboulle:
you might also reject this with the stipulation that the developer is responsible for setting a Docker-appropriate type (e.g. `container_file_t` on Fedora) on her source directory.. dunno.

cyphar:
I'd prefer `:z` or `:Z`, to be honest. While it's more expensive, disabling SELinux labeling entirely doesn't seem like a good idea (I'd prefer to test in as many environments as possible).

cyphar:
Or actually, maybe we just shouldn't use volumes for the source code? The reason for having `-v` is so that `make shell` is useful, but I use that so rarely maybe we should just drop `-v` entirely.

jonboulle:
> I'd prefer :z or :Z, to be honest.

IMO it's a bit gross for a `make` invocation to be munging attributes of the source directory :man_shrugging: 

> Or actually, maybe we just shouldn't use volumes for the source code?

and do what instead?

cyphar:
@jonboulle Just do `COPY`. This will result in the source code being owned by the correct label.

jonboulle:
Do you want me to drop the shell target..?

cyphar:
Ah wait, the CI might break now because we store the coverage profiles for each run between containers in the source directory... The fix is probably going to be even more involved than I'd like (we have to change how the current `COVERAGE` magic works). If you want, I can take a crack at this one since it requires touching a bunch of different CI bits.

jonboulle:
ick, this results in other SELinux issues - denied `unmount`/`remount`s.

cyphar:
Alright, I guess that the original patch is probably the most ideal way of handling it. That's a bit of a pain. 😢

cyphar:
[FWIW, I am planning on dropping the `remount` code in the future, but that'll have to wait for some other refactors I've got sitting around on a branch somewhere.]

jonboulle:
ok, original posted at #206

cyphar:
Dropping in favour of #206.

PR #205, opened 2017-10-24
Makefile: add option to specify individual test(s)

jonboulle:
Running all the integration tests can be a bit tedious when a developer
is iterating on a particular subcommand. This adds the ability to
specify a subset of tests by passing the `TESTS=` argument to the
`test-integration` and `local-test-integration` Make targets. This
argument expects the name of one or more tests, each of which is
expected to match a .bats file in the `tests/` directory. If unset, we
continue to default to running all tests.

Examples:
  $ make TESTS="gc" test-integration
  $ make TESTS="unpack repack" local-test-integration
  $ make test-integration

### DIFF
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -33,7 +33,15 @@ if [ "$COVER" -eq 1 ]; then
 fi
 
 # Run the tests and collate the results.
-bats -t $ROOT/test/*.bats
+tests=()
+if [[ -z "$TESTS" ]]; then
+	tests=($ROOT/test/*.bats)
+else
+	for f in $TESTS; do
+		tests+=("$ROOT/test/$f.bats")
+	done
+fi
+bats -t ${tests[*]}
 if [ "$COVER" -eq 1 ]; then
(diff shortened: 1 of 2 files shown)


cyphar:
Supersedes #147.

jonboulle:
One day I will learn to read..

hack/test-integration.sh
```diff
@@ -33,7 +33,15 @@ if [ "$COVER" -eq 1 ]; then
 fi
 
 # Run the tests and collate the results.
-bats -t $ROOT/test/*.bats
+tests=""
+if [[ -z "${TESTS}" ]]; then
+  tests=${ROOT}/test/*.bats
+else
+  for f in ${TESTS:-*}; do
```
cyphar: The `:-*` is unnecessary here. Also, if you could use an array here rather than a string that'd be great. Something like:

```
tests=()
if [ -z "$TESTS" ]; then
  tests=( $ROOT/test/*.bats )
else
  for f in $TESTS; do
    tests+=( "$ROOT/test/$f.bats")
  done
fi
```

Also, tabs > spaces. :wink:

jonboulle:
updated

cyphar:
LGTM.

PR #206, opened 2017-10-24
Makefile: disable SELinux for tests in Docker

jonboulle:
Re-do of #203. As discussed there, there are outstanding SELinux issues with the tests even if we avoid the labelling one.

With the current Makefile, the Docker variants of the tests don't work
out of the box on SELinux-enabled systems because the volume mount
(namely, the source directory) isn't labelled correctly. This patch
simply disables SELinux labelling for these particular containers,
as the easiest way to mitigate this issue. (This seems preferable to
using the `:z` or `:Z` option to the Docker volume mount, which would
forceably relabel the directory every time the Docker command is run.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ CMD := ${PROJECT}/cmd/umoci
 
 # We use Docker because Go is just horrific to deal with.
 UMOCI_IMAGE := umoci_dev
+DOCKER_RUN := docker run --rm -it --security-opt label:disable -v ${PWD}:/go/src/${PROJECT}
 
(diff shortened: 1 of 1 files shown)


cyphar:
Needs a rebase.

jonboulle:
done

cyphar:
LGTM.

cyphar:
Travis failure is quite odd. It looks like the `skopeo copy docker://fedora:latest ...` is blocking indefinitely for some reason.

cyphar:
LGTM.

PR #207, opened 2017-10-24
test: change oci-image-validate reference

jonboulle:
As of
https://github.com/opencontainers/image-tools/commit/1423c5e01935372d9f069e158f11a0f1b43df221,
oci-image-tools moved the individual commands into subcommands (i.e.
`oci-image-validate` became `oci-image-tool validate`. The umoci
development Dockerfile installs a very old version of the OCI image
tools [1][1] before this merge happened; and although it installs a
specific (newer) version of the image tools "over the top" [2][2],
during the tests we were specifically calling `oci-image-validate` which
would invoke the old version.

[1]: https://github.com/openSUSE/umoci/blob/80f8d32c23772c3d9f20d02f82fd6a60186c0e92/Dockerfile#L35
[2]: https://github.com/openSUSE/umoci/blob/80f8d32c23772c3d9f20d02f82fd6a60186c0e92/Dockerfile#L67

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM opensuse/amd64:42.2
+FROM opensuse/amd64:42.3
 MAINTAINER "Aleksa Sarai <asarai@suse.com>"
 
 # We have to use out-of-tree repos because several packages haven't been merged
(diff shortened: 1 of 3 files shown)


jonboulle:
Caught this because tests that I was expecting to fail during the image validation weren't doing so :-(

@cyphar is there actually any point in installing the `oci-image-tools` RPM?

cyphar:
@jonboulle I'm going to drop the re-compilation of the `oci-*-tools` in the near future, once some upstream bugs are fixed and we can just use the RPMs. One of those bugs has been fixed recently, so I might just be able to do the updates now...

cyphar:
@jonboulle I've rebased this, and removed all of the additional reinstall code -- so we're only using the openSUSE packages.

cyphar:
LGTM.

PR #208, opened 2017-10-25
*: switch entirely away from system.Mknod

cyphar:
Since golang.org/x/sys/unix already has all the utilities we need for
messing with dev_t values we can just remove our wrappers entirely. This
also allows us to remove all of the special-casing we have for golint.

Follow-up: 722012370cda ("pkg: system: use Major, Minor and Mkdev from x/sys/unix")


cyphar:
LGTM.

PR #209, opened 2017-10-25
ci: small fixes

cyphar:
Remove un-used OBS projects from Dockerfile, as well as updating the
version of the openSUSE image we test.


cyphar:
LGTM.

PR #212, opened 2017-10-28
 layer: improve warnings in rootless mode

cyphar:
It appears that golang.org/x/sys/unix doesn't add path information to
errors (and only return the syscall.Errno). While this is fine, it
caused rootless warnings to be needlessly confusing. In addition, the
warnings looked too scary before. Hopefully they seem less worrying now.

In addition, add a new warning when we fake a device inode. Most of the
container images I've tested don't appear to contain them, so the
warning isn't too spammy (and also it doesn't look as worrying
hopefully).

/cc @davidcassany 


cyphar:
LGTM.

PR #219, opened 2017-12-13
API draft

tych0:
Hi,

I'm interested in using umoci from a go program; rather than do all the fork+exec stuff, I hoisted some of the code out to an API. I only hoisted a small portion of the code, basically what I need to accomplish what I want.

Thoughts welcome.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,0 +1,195 @@
+package umoci
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/openSUSE/umoci/oci/cas"
+	"github.com/openSUSE/umoci/oci/cas/dir"
+	"github.com/openSUSE/umoci/oci/casext"
+	igen "github.com/openSUSE/umoci/oci/config/generate"
+	"github.com/opencontainers/go-digest"
+	imeta "github.com/opencontainers/image-spec/specs-go"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+)
+
+type Layout struct {
+	engine cas.Engine
+	ext    casext.Engine
+}
+
+func OpenLayout(imagePath string) (*Layout, error) {
+	l := &Layout{}
+	var err error
+	// Get a reference to the CAS.
+	l.engine, err = dir.Open(imagePath)
+	if err != nil {
+		return nil, errors.Wrap(err, "open CAS")
+	}
+
+	l.ext = casext.NewEngine(l.engine)
+
+	return l, nil
+}
+
+func CreateLayout(imagePath string) (*Layout, error) {
+	err := dir.Create(imagePath)
+	if err != nil {
+		return nil, err
+	}
+
+	return OpenLayout(imagePath)
+}
+
+func (l *Layout) resolve(fromName string) (ispec.Descriptor, error) {
+	descriptorPaths, err := l.ext.ResolveReference(context.Background(), fromName)
+	if err != nil {
+		return ispec.Descriptor{}, err
+	}
+
+	if err != nil {
+		return ispec.Descriptor{}, errors.Wrap(err, "get descriptor")
+	}
+	if len(descriptorPaths) == 0 {
+		return ispec.Descriptor{}, errors.Errorf("tag not found: %s", fromName)
+	}
+	if len(descriptorPaths) != 1 {
+		// TODO: Handle this more nicely.
+		return ispec.Descriptor{}, errors.Errorf("tag is ambiguous: %s", fromName)
+	}
+	descriptor := descriptorPaths[0].Descriptor()
+	return descriptor, nil
+}
+
+func (l *Layout) Tag(from string, to string) error {
+	descriptor, err := l.resolve(from)
+	if err != nil {
+		return err
+	}
+
+	if err := l.ext.UpdateReference(context.Background(), to, descriptor); err != nil {
+		return errors.Wrap(err, "put reference")
+	}
+
+	return nil
+}
+
+func (l *Layout) PutBlob(b io.Reader) (Layer, error) {
+	// This unpacking is a little awkward, but I don't know how to work
+	// around the vendoring of go-digest.
+	digest, size, err := l.engine.PutBlob(context.Background(), b)
+	if err != nil {
+		return Layer{}, err
+	}
+
+	return Layer{Hash: string(digest), Size: size}, nil
+}
+
+type Layer struct {
+	Hash string
+	Size int64
+}
+
+func (l Layer) ToDigest() (digest.Digest, error) {
+	return digest.Parse(l.Hash)
+}
+
+func (l *Layout) NewImage(tagName string, g *igen.Generator, layers []Layer, mediaType string) error {
+	layerDescriptors := []ispec.Descriptor{}
+	for _, l := range layers {
+		d, err := digest.Parse(l.Hash)
+		if err != nil {
+			return err
+		}
+
+		layerDescriptors = append(layerDescriptors, ispec.Descriptor{
+			MediaType: mediaType,
+			Digest:    d,
+			Size:      l.Size,
+		})
+	}
+
+	// Update config and create a new blob for it.
+	config := g.Image()
+	configDigest, configSize, err := l.ext.PutBlobJSON(context.Background(), config)
+	if err != nil {
+		return errors.Wrap(err, "put config blob")
+	}
+
+	// Create a new manifest that just points to the config and has an
+	// empty layer set. FIXME: Implement ManifestList support.
+	manifest := ispec.Manifest{
+		Versioned: imeta.Versioned{
+			SchemaVersion: 2, // FIXME: This is hardcoded at the moment.
+		},
+		Config: ispec.Descriptor{
+			MediaType: ispec.MediaTypeImageConfig,
+			Digest:    configDigest,
+			Size:      configSize,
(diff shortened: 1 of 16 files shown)


cyphar:
Heh, that's interesting -- @hallyn was working on this as well.

tych0:
Yep, we're working on the same project :).

I'll add some docs, seems like that's what travis is complaining about.

hallyn:
Yup - I had started my own (in my github tree, though not all commits were pushed) but I'm pretty sure Tycho started from scratch.  We've talked though and both share the goal of hiding some of the large number of confusing different types that currently need to be used.  We're hoping umoci will offer (and for it's tools, use) a clean API that can be used with just a few types with simple easy to understand API.  Browsing the patchset here that seems to be happening, so cool.

I'll take a closer look, both at this set, and at the new user :)

tych0:
Ping, any thoughts on this?

hallyn:
The patches look good to me.  But before we have users, let's talk about the API.

Right now we have 2 constructors - OpenLayout() and CreateLayout() - and the public methods:

Tag
PutBlob
ToDigest
NewImage
ListTags
Close
LayersForTag

A few questions:

1. do we want to support multiple simultaneous users of a layoutdir (which implies no caching of any data)?
2. Do we want to stick to a single entrypoint (i.e. make OpenLayout optionally create?  Probably not worth it)
3. Should Image by a type of its own?  Blob actions could go through it?

tych0:
r.e. 1, I think this should work as is, if I understand correctly.
2, I just copied what the umoci binary already does here, which is complain if you specify `new` and the thing exists, or `tag list` and it doesn't. I don't have a strong opinion either way, though.
3, I'm actually currently refactoring Layer into the type Blob (since that's what it really is, basically an alias for ispec.Descriptor since we can't use that); I don't think blob operations should go through it, though, since blobs can be re-used by multiple images.

hallyn:
@cyphar,

how do you feel about this approach?  We're starting to use this now in github.com/anuvu/stacker, so we either need an API agreed upon soon, or we can live off of a forked umoci for awhile.

cyphar:
I'm going to review this properly later this week (likely on Thursday or Friday). From a first look I have a few comments, mainly related to how things are exposed through the API.

* I don't think `LayersForTag` makes as much sense as providing a metadata object that contains that information, so you'd do `Stat(...)` and then return a struct which contains a list of layer digests (or layer descriptors). This would be able to be used to make a nicer `umoci stat` CLI API in the future. Of course this could be refined in the future -- it's up to you if you want to do it now.

* Why are we not using `go-digest` when returning values? Seems odd.

* There's no need to have a `casext` and a `cas`. `casext` is an extension of `cas` (it embeds the object so it can have the same interface -- though in future I might make it just embed the interface).

* (This is more for future work IMO) If we're going to have a `resolve` API I would at least like to have a structured error for ambiguous resolution problems -- it could be unused for now but we could use it to allow for interactive resolution code.

* Your `Layer` object doesn't specify some information about the layers like is the layer compressed, and whether the layer is non-distributable. Personally I think we should just return a `Descriptor` here to avoid duplication of objects.

Now to answer these questions:

>  1. Do we want to support multiple simultaneous users of a layoutdir (which implies no caching of any data)?

Yes, and `umoci` already supports it (at least in principle).

> 2. Do we want to stick to a single entrypoint (i.e. make OpenLayout optionally create? Probably not worth it)

I think that having separate open and create entrypoints make sense.

> Should Image by a type of its own? Blob actions could go through it?

I would say that we probably should have Image be its own type, but I'd need to mess around with an API like that before deciding if it's actually easier to use.

--- 

Finally, sorry for not working on `umoci` as much as I was previously, I've been quite busy with a few other bits and pieces at the moment related to Kubernetes. I will make some more time for `umoci` maintainence.

tych0:
On Wed, Dec 20, 2017 at 05:39:39AM +0000, Aleksa Sarai wrote:
> I'm going to review this properly later this week (likely on Thursday or Friday). From a first look I have a few comments, mainly related to how things are exposed through the API.
>
> * I don't think `LayersForTag` makes as much sense as providing a metadata object that contains that information, so you'd do `Stat(...)` and then return a struct which contains a list of layer digests (or layer descriptors). This would be able to be used to make a nicer `umoci stat` CLI API in the future. Of course this could be refined in the future -- it's up to you if you want to do it now.
>
> * Why are we not using `go-digest` when returning values? Seems odd.
>
> * Your `Layer` object doesn't specify some information about the layers like is the layer compressed, and whether the layer is non-distributable. Personally I think we should just return a `Descriptor` here to avoid duplication of objects.

These are all essentially the same complaint, and I completely agree.
Unfortunately, this is all due to golang vendoring terrible-ness.

If we use a descriptor directly, that makes all users of the umoci API have to
invent their own wrappers, because of go's vendoring. For example, if I write a
function in my library:

func ReturnDescriptor() ispec.Descriptor {
        return umoci.ReturnDescriptor()
}

I get:

./storage.go:338:31: cannot use umoci.ReturnDescriptor() (type "github.com/openSUSE/umoci/vendor/github.com/opencontainers/image-spec/specs-go/v1".Descriptor) as type "github.com/opencontainers/image-spec/specs-go/v1".Descriptor in return argument

This also means (and is probably a bigger problem) that the umoci API can't
ever take an ispec.Descriptor as an argument, because external users have no
way to create that type. You can't even import the vendored version:

storage.go:15:2: use of vendored package not allowed

So, IMO it's best to define a wrapper type that is identical (or better yet,
just stop vendoring the image spec). Layer (or as I've renamed it in my local
branch, "Blob") is basically just a descriptor.

> * There's no need to have a `casext` and a `cas`. `casext` is an extension of `cas` (it embeds the object so it can have the same interface -- though in future I might make it just embed the interface).

Thanks, will fix.

tych0:
On Wed, Dec 20, 2017 at 07:48:14PM +0000, Aleksa Sarai wrote:
> > ./storage.go:338:31: cannot use umoci.ReturnDescriptor() (type "github.com/openSUSE/umoci/vendor/github.com/opencontainers/image-spec/specs-go/v1".Descriptor) as type "github.com/opencontainers/image-spec/specs-go/v1".Descriptor in return argument
> 
> Ah, shit. I completely forgot about that. I think the way that you "fix" this problem is that you have to flatten the vendor tree in the top-level project (so you vendor it and then) of whatever user you have of `umoci`. The problem is not caused by third-party repos, but just that Go cannot reconcile the types of two different vendor trees being the same (even if the code is identical).

Heh, I guess it's a matter of opinion on whose fault it is, but yes :)

> Another possible option is to switch `umoci` to using `gopkg.in` (like LXD does). It will require quite a bit of work (especially for the packaging side for us in openSUSE) but that could be a solution as well.

LXD doesn't use vendoring at all, but go-lxc does use gopkg.in. As you
say, it's probably more work than just using vendoring in umoci.

Anyway, I'll try vendoring in stacker. It's ugly, but so is vendoring
in general :)

cyphar:
On 2017-12-20, Tycho Andersen <reply@reply.github.com> wrote:
> LXD doesn't use vendoring at all, but go-lxc does use gopkg.in. As you
> say, it's probably more work than just using vendoring in umoci.
> 
> Anyway, I'll try vendoring in stacker. It's ugly, but so is vendoring
> in general :)

Yeah, I completely agree. Go punted on all the wrong things in the worst
way possible. This is one of the reasons I've started writing things in
Rust. While it also has it's packaging issues (the management system is
based on JavaScript) it's at least somewhat sane.

-- 
Aleksa Sarai
Senior Software Engineer (Containers)
SUSE Linux GmbH
<https://www.cyphar.com/>

hallyn:
@cyphar 
> This is one of the reasons I've started writing things in Rust. 

Does this include opencontainers related tools?

cyphar:
On 2017-12-22, Serge Hallyn <reply@reply.github.com> wrote:
> > This is one of the reasons I've started writing things in Rust.
> Does this include opencontainers related tools?

Unfortunately not at the moment. We have discussed rewriting runc in
Rust a few times, and I think most of the maintainers an on-board with
that idea. I do wish that I'd written umoci in Rust originally, but I'm
not sure how much work it would be to rewrite it in Rust (I checked and
there don't seem to be any good "tar" crates, which is necessary for
umoci to be able to work). But we'll see.

It's a massive shame that the container community seems to have settled
on Go. It's the least suitable language for that purpose in my opinion.

-- 
Aleksa Sarai
Senior Software Engineer (Containers)
SUSE Linux GmbH
<https://www.cyphar.com/>

cyphar:
@tych0 This is looking pretty good (aside from needing a rebase, some squashing and probably some docs down the line). I was considering carrying this if you like (some of the API decisions around `Lookup*` are a little odd and might need some cleaning -- and I'm not sure we should burn two people's time over my OCD :smile_cat:).

You mention in the description that you need this for a Go program -- is that program hosted somewhere? Can I take a look at it to see how you're using the API?

tych0:
Actually I have a better design for this... where we don't replicate all the Lookup* stuff and just expose the public Engine. Unfortunately I have had zero time to work on it in the last few months, but hopefully sometime soon. Anyway, you can close this if you want if you're cleaning up, in favor of that.

tych0:
Yes, the new design would just leave it up to consumers what to do with multiple tags, so I think it would be fine. I'm not sure when I'll get to it exactly, it's been on my list for at least two months, but here we are :)

tych0:
I'm gonna close this in favor of the version above.

PR #220, opened 2017-12-14
don't always gunzip layers

tych0:
Layers can be stored in tar format alone, so let's not force gunzip.

### DIFF
--- a/oci/layer/unpack.go
+++ b/oci/layer/unpack.go
@@ -82,6 +82,10 @@ func isLayerType(mediaType string) bool {
 		mediaType == ispec.MediaTypeImageLayerGzip || mediaType == ispec.MediaTypeImageLayerNonDistributableGzip
 }
 
+func needsGunzip(mediaType string) bool {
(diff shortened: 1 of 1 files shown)


oci/layer/unpack.go
```diff
@@ -226,7 +234,7 @@ func UnpackManifest(ctx context.Context, engine cas.Engine, bundle string, manif
 		// in the layer.
 		_, _ = io.Copy(ioutil.Discard, layer)
 		// XXX: Is it possible this breaks in the error path?
-		layerGzip.Close()
+		layerRaw.Close()
```
cyphar: This should be `layerData`, assuming you wanted to have the same semantics.

tych0: Arg, yeah, thanks. Just pushed a fix.

cyphar:
LGTM.

cyphar:
Sorry, something's wrong with my `lgtm.co` instance. I'm looking into it...

cyphar:
LGTM.

PR #227, opened 2018-03-02
oci: layer: add user.rootlesscontainers support

cyphar:
Add the ability for a rootless user to extract a layer while still
providing some information for tools like PRoot that specify what the
"real" owner of a file is. This support effectively just boils down to
adding some more cases to {,un}mapHeader.

Cc: @AkihiroSuda


pkg/rootlesscontainers-proto/README.md
```diff
@@ -0,0 +1,14 @@
...
+what this protobuf blob means. Effectively this is used in rootless mode for
+the purposes of emulating `chown(2)` within a rootless container (using
+something like `remainroot` or `PRoot`).
+
+The protobuf itself comes from [the website repo][protobuf-source], and this
+package's sources can be regenerated using `go generate` (assuming you have
+`protoc` set up).
+
+[rootlesscontaine.rs]: https://rootlesscontaine.rs/
+[protobuf-source]: https://github.com/cyphar/rootlesscontaine.rs/blob/master/proto/rootlesscontainers.proto
```
AkihiroSuda: I think you should put the commit hash here

oci/layer/utils.go
```diff
@@ -45,24 +48,49 @@ type MapOptions struct {
...
+		}
+
+		// If the payload isn't uint32(-1) we apply it. The xattr includes the
+		// *in-container* owner so we don't want to map it.
+		if uid := payload.GetUid(); uid != rootlesscontainers.NoopId {
+			newUid = int(uid)
+		}
+		if gid := payload.GetGid(); gid != rootlesscontainers.NoopId {
+			newGid = int(gid)
+		}
```
AkihiroSuda: I think `hdr.Xattrs[rootlesscontainers.Keyname]` should be removed here

cyphar: You're right -- the code used to do that but I must've lost it in a refactor. The tests I'm working on would've picked this up (for what it's worth).

oci/layer/utils.go
```diff
@@ -71,10 +99,57 @@ func mapHeader(hdr *tar.Header, mapOptions MapOptions) error {
...
+		// Don't add the xattr if the owner isn't just (0, 0) because that's a
+		// waste of space.
+		if !rootlesscontainers.IsDefault(payload) {
+			valueBytes, err := proto.Marshal(&payload)
+			if err != nil {
+				return errors.Wrap(err, "marshal rootlesscontainers payload")
+			}
+			// While the payload is almost certainly not UTF-8, Go strings can
+			// actually be arbitrary bytes (in case you didn't know this and
+			// were confused like me when this worked).
```
AkihiroSuda: I think we should cite https://blog.golang.org/strings

pkg/rootlesscontainers-proto/README.md
AkihiroSuda: > What would be even more ideal is if the .proto file was accessible from https://rootlesscontaine.rs/proto/rootlesscontainers.proto 

It is fine if the server can add a header like this:
```
// Autogenerated from https://github.com/cyphar/rootlesscontaine.rs/blob/ae6660f7f2e8db1d356272446caab183a44a3bd7/proto/rootlesscontainers.proto
// DO NOT EDIT MANUALLY
```

cyphar:
Alright @AkihiroSuda -- this should be ready. Do you wanna take a quick look before I merge?

cyphar:
@AkihiroSuda That's fair enough -- I think that the best way of doing this would be to have a Go repository that we vendor rather than generating it within `umoci`. If you don't mind, I will do this in a follow-up (I also don't really like the way I did https://rootlesscontaine.rs/proto/rootlesscontainers.proto -- it's really odd to have a protobuf spec that is meant to be versioned in the middle of an unversioned website).

cyphar:
LGTM.

PR #229, opened 2018-03-09
oci: tar extract: full opaque whiteout support

### LINKED ISSUE #224: make whiteout handling spec-compliant
At the moment, our handling of whiteout files in layers isn't correct. According to the spec, whiteouts need to be applied *before* anything else in the archive. However, this leads us to a dilemma. The only way to figure out what items are present in an archive is to read the whole thing into memory (bad) or to require that the archive be seekable (also bad) in order for us to scan over the archive twice (once to collect the whiteouts, and the second time to apply the legitimately "new" files).

But in short we need to implement a few key things:

* Apply whiteouts before all other entries.
* Add support for `.wh..wh..opq`.
* Refuse to create a layer if a (real) file has `.wh.` as a prefix.

cyphar:
Add full opaque whiteout support, including the handling of new files
present in the upper layer (and thus should not be removed). The code
implementing this is fairly fragile and has to deal with a lot of really
annoying edge cases, so many new tests were added for this portion.

There are a few semantics differences found between unpriv.Walk and
filepath.Walk by this patch (especially when it comes to handling ENOENT
during the walk). These semantics have been fixed in the WalkFunc used,
but probably deserve a refactor in future.

Closes #224 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   fork][as-proot-fork] (though we do not test its interoperability at the
   moment) as both tools use [the same protobuf
   specification][rootlesscontainers-proto]. openSUSE/umoci#227
+- `umoci unpack` now has support for opaque whiteouts (whiteouts which remove
+  all children of a directory in the lower layer), though `umoci repack` does
+  not currently have support for generating them. While this is technically a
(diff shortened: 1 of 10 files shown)


cyphar:
We can handle generating opaque whiteouts in the future (because we technically shouldn't be generating them).

cyphar:
LGTM.

cyphar:
LGTM.

PR #230, opened 2018-03-09
oci: tar extract: enable some rootless tricks in user namespace

### LINKED ISSUE #171: umoci-unpack fails under a user namespace
This is because we currently bail on `mknod` failures (and so on). We really should enable the "rootless" tricks if we are currently in a user namespace (which we can check like other projects do, with `setgroups` and `uid_map`).

Reported-by: @hallyn

cyphar:
Some rootless tricks we do (such as faking block devices) are useful in
user namespaces, while others (ignoring Lchown) are not. In the future
the way we handle this will probably change (especially with the advent
of seccomp-userspace traps which would allow for mknod(2) emulation in
userspace -- removing the privilege problem), but currently we enable
some tricks if we are detected as being in a user namespace.

It should be noted that user namespace checks are very fickle. There
isn't any way to guarantee you are not in a user namespace without
trying something like mknod(2) and noticing it fails when it shouldn't
(and even that could be blocked by other things). So the check we do
(see if /proc/self/uid_map has the same value as the root user
namespace) is 99% accurate though it can be tricked quite trivially.

Closes #171
Reported-by: Serge Hallyn <serge.hallyn@ubuntu.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   not currently have support for generating them. While this is technically a
   spec requirement, through testing we've never encountered an actual user of
   these whiteouts. openSUSE/umoci#224 openSUSE/umoci#229
+- `umoci unpack` will now use some rootless tricks inside user namespaces for
+  operations that are known to fail (such as `mknod(2)`) while other operations
+  will be carried out as normal (such as `lchown(2)`). It should be noted that
+  the `/proc/self/uid_map` checking we do can be tricked into not detecting
+  user namespaces, but you would need to be trying to break it on purpose.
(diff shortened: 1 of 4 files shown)


cyphar:
LGTM. This is the only thing blocking the next release.

hallyn:
Seems to work. Inside an lxd container, I pulled docker://ubuntu, unpacked it, added a file, and repacked it.  The original layers persisted with a new tarball with just the new file, just as it should be.

Thanks.

cyphar:
Awesome. I haven't been able to add tests for this yet (you can't tell Docker to run a single container in a user namespace, and I'd have to write some wrapper around `unshare` and `newuidmap`/`newgidmap` to make it work) but I will do that in the future.

PR #232, opened 2018-03-14
oci: tar extract: handle out-of-order regular whiteouts

### LINKED ISSUE #224: make whiteout handling spec-compliant

cyphar:
08da5c6d0fd9 ("oci: tar extract: full opaque whiteout support") added
handling for opaque whiteouts such that out-of-order whiteouts were
still handled correctly. However, this new handling wasn't extended to
regular whiteouts (which have effectively the same semantics).

Correct this by refactoring out the removal function we crafted for
opaque whiteouts and just use it for all whiteouts.

Closes #224 (again).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- `umoci unpack` now handles out-of-order regular whiteouts correctly (though
+  this ordering is not recommended by the spec -- nor is it required). This is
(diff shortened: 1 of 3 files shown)


cyphar:
LGTM.

PR #237, opened 2018-04-17
add `umoci insert` command

tych0:
There are cases where you only want to insert files into an image. Right
now the procedure is: unpack the image, add the files, and repack the
image. However, the OCI format lends itself exactly to this kind of
addition -- we can just add a new layer with the files we wish to insert.
Additional motivation is in the case of large layers, we can avoid a lot of
i/o, CPU time to compress, and disk space requirements.

So, let's add a command to generate this new import layer.

### DIFF
--- a/cmd/umoci/insert.go
+++ b/cmd/umoci/insert.go
@@ -1,0 +1,140 @@
+/*
+ * umoci: Umoci Modifies Open Containers' Images
+ * Copyright (C) 2018 Cisco
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"context"
+	"time"
+
+	"github.com/apex/log"
+	"github.com/openSUSE/umoci/mutate"
+	"github.com/openSUSE/umoci/oci/cas/dir"
+	"github.com/openSUSE/umoci/oci/casext"
+	igen "github.com/openSUSE/umoci/oci/config/generate"
+	"github.com/openSUSE/umoci/oci/layer"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
+)
+
+var insertCommand = uxHistory(cli.Command{
+	Name:  "insert",
+	Usage: "insert a file into an OCI image without unpacking/repacking it",
+	ArgsUsage: `--image <image-path>[:<tag>] <file> <path>
+
+Where "<image-path>" is the path to the OCI image, "<tag>" is the name of the
+tag that the content wil be inserted into (if not specified, defaults to
+"latest"), "<file>" is the file or folder to insert, and "<path>" is the prefix
+inside the image to insert it into.`,
+
+	Category: "image",
+
+	Action: insert,
+
+	Before: func(ctx *cli.Context) error {
+		if ctx.NArg() != 2 {
+			return errors.Errorf("invalid number of positional arguments: expected <file> and <path>")
+		}
+		if ctx.Args()[0] == "" {
+			return errors.Errorf("path cannot be empty")
+		}
+		if ctx.Args()[1] == "" {
+			return errors.Errorf("path cannot be empty")
+		}
+		return nil
+	},
+})
+
+func insert(ctx *cli.Context) error {
+	imagePath := ctx.App.Metadata["--image-path"].(string)
+	tagName := ctx.App.Metadata["--image-tag"].(string)
+
+	// Get a reference to the CAS.
+	engine, err := dir.Open(imagePath)
+	if err != nil {
+		return errors.Wrap(err, "open CAS")
+	}
+	engineExt := casext.NewEngine(engine)
+	defer engine.Close()
+
+	descriptorPaths, err := engineExt.ResolveReference(context.Background(), tagName)
+	if err != nil {
+		return errors.Wrap(err, "get descriptor")
+	}
+	if len(descriptorPaths) == 0 {
+		return errors.Errorf("tag not found: %s", tagName)
+	}
+	if len(descriptorPaths) != 1 {
+		// TODO: Handle this more nicely.
+		return errors.Errorf("tag is ambiguous: %s", tagName)
+	}
+
+	// Create the mutator.
+	mutator, err := mutate.New(engine, descriptorPaths[0])
+	if err != nil {
+		return errors.Wrap(err, "create mutator for base image")
+	}
+
+	// TODO: add some way to specify these from the cli
+	reader := layer.GenerateInsertLayer(ctx.Args()[0], ctx.Args()[1], nil)
+	defer reader.Close()
+
+	created := time.Now()
+	history := ispec.History{
+		Comment:    "",
+		Created:    &created,
+		CreatedBy:  "umoci insert", // XXX: Should we append argv to this?
+		EmptyLayer: false,
+	}
+
+	if val, ok := ctx.App.Metadata["--history.author"]; ok {
+		history.Author = val.(string)
+	}
+	if val, ok := ctx.App.Metadata["--history.comment"]; ok {
+		history.Comment = val.(string)
+	}
+	if val, ok := ctx.App.Metadata["--history.created"]; ok {
+		created, err := time.Parse(igen.ISO8601, val.(string))
+		if err != nil {
+			return errors.Wrap(err, "parsing --history.created")
+		}
+		history.Created = &created
+	}
+	if val, ok := ctx.App.Metadata["--history.created_by"]; ok {
+		history.CreatedBy = val.(string)
+	}
+
+	// TODO: We should add a flag to allow for a new layer to be made
+	//       non-distributable.
+	if err := mutator.Add(context.Background(), reader, history); err != nil {
+		return errors.Wrap(err, "add diff layer")
+	}
+
+	newDescriptorPath, err := mutator.Commit(context.Background())
+	if err != nil {
+		return errors.Wrap(err, "commit mutated image")
+	}
+
+	log.Infof("new image manifest created: %s->%s", newDescriptorPath.Root().Digest, newDescriptorPath.Descriptor().Digest)
+
+	if err := engineExt.UpdateReference(context.Background(), tagName, newDescriptorPath.Root()); err != nil {
+		return errors.Wrap(err, "add new tag")
+	}
+	log.Infof("updated tag for image manifest: %s", tagName)
+	return nil
+}
--- a/cmd/umoci/insert.go
+++ b/cmd/umoci/insert.go
@@ -33,7 +33,7 @@ import (
 	"github.com/urfave/cli"
 )
 
-var insertCommand = uxHistory(cli.Command{
+var insertCommand = uxRemap(uxHistory(cli.Command{
 	Name:  "insert",
 	Usage: "insert a file into an OCI image without unpacking/repacking it",
 	ArgsUsage: `--image <image-path>[:<tag>] <file> <path>
@@ -69,7 +69,7 @@ For example:
 
 		return nil
 	},
-})
+}))
 
 func insert(ctx *cli.Context) error {
 	imagePath := ctx.App.Metadata["--image-path"].(string)
@@ -104,8 +104,16 @@ func insert(ctx *cli.Context) error {
 	insertFile := ctx.App.Metadata["insertFile"].(string)
 	insertPath := ctx.App.Metadata["insertPath"].(string)
 
-	// TODO: add some way to specify these from the cli
-	reader := layer.GenerateInsertLayer(insertFile, insertPath, nil)
+	var meta UmociMeta
+	meta.Version = UmociMetaVersion
+
+	// Parse and set up the mapping options.
+	err = parseIdmapOptions(&meta, ctx)
+	if err != nil {
+		return err
+	}
+
+	reader := layer.GenerateInsertLayer(insertFile, insertPath, &meta.MapOptions)
 	defer reader.Close()
 
 	created := time.Now()
--- a/cmd/umoci/insert.go
+++ b/cmd/umoci/insert.go
@@ -1,5 +1,6 @@
 /*
  * umoci: Umoci Modifies Open Containers' Images
+ * Copyright (C) 2016, 2017, 2018 SUSE LLC.
  * Copyright (C) 2018 Cisco
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/cmd/umoci/insert.go
+++ b/cmd/umoci/insert.go
@@ -40,8 +40,14 @@ var insertCommand = uxHistory(cli.Command{
 
 Where "<image-path>" is the path to the OCI image, "<tag>" is the name of the
 tag that the content wil be inserted into (if not specified, defaults to
-"latest"), "<file>" is the file or folder to insert, and "<path>" is the prefix
-inside the image to insert it into.`,
+"latest"), "<file>" is the file or folder to insert, and "<path>" is the full
+name of the path to that the file should be inserted at. Insert is
+automatically recursive if the source is a directory.
+
+For example:
+	umoci insert --image oci:foo mybinary /usr/bin/mybinary
+	umoci insert --image oci:foo myconfigdir /etc/myconfigdir
+`,
 
 	Category: "image",
 
@@ -52,11 +58,15 @@ inside the image to insert it into.`,
 			return errors.Errorf("invalid number of positional arguments: expected <file> and <path>")
 		}
 		if ctx.Args()[0] == "" {
-			return errors.Errorf("path cannot be empty")
+			return errors.Errorf("<file> cannot be empty")
 		}
+		ctx.App.Metadata["insertFile"] = ctx.Args()[0]
+
 		if ctx.Args()[1] == "" {
-			return errors.Errorf("path cannot be empty")
+			return errors.Errorf("<path> cannot be empty")
 		}
+		ctx.App.Metadata["insertPath"] = ctx.Args()[1]
+
 		return nil
 	},
 })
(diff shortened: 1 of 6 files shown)


oci/layer/generate.go
```diff
@@ -97,3 +100,44 @@ func GenerateLayer(path string, deltas []mtree.InodeDelta, opt *MapOptions) (io.
...
+	go func() {
+		var err error
+
+		defer func() {
+			writer.CloseWithError(errors.Wrap(err, "generate layer"))
+		}()
+
+		tg := newTarGenerator(writer, mapOptions)
+
+		err = filepath.Walk(root, func(curPath string, info os.FileInfo, err error) error {
```
cyphar: I think this should be using `unpriv.Walk` (with the regular checks for which `fs` we should use). One of the things people want is for `umoci` to use it's magic `CAP_DAC_OVERRIDE` emulation to be able to handle annoying paths.

And once we add that, we'll need another integration test to make sure that the path modes are included correctly.

tych0:
Hey, sorry for the delay, I updated it and added a stat test.

test/insert.bats
```diff
@@ -0,0 +1,52 @@
...
+@test "umoci insert" {
+	image-verify "${IMAGE}"
+
+	# fail with too few arguments
+	umoci insert --image "${IMAGE}:${TAG}"
+	[ "$status" -ne 0 ]
+
+	# ...and too many
+	umoci insert --image "${IMAGE}:${TAG}" asdf 123 456
+	[ "$status" -ne 0 ]
```
cyphar: I prefer this to be in a separate test (to match the other "are the arguments handled correctly" tests).

test/insert.bats
```diff
@@ -0,0 +1,52 @@
...
+
+	# do an insert
+	umoci insert --image "${IMAGE}:${TAG}" "${ROOT}/test/insert.bats" /tester
+	[ "$status" -eq 0 ]
+
+	umoci insert --image "${IMAGE}:${TAG}" "${ROOT}/umoci" /tester
+	[ "$status" -eq 0 ]
```
cyphar: I would also recommend having a rootfs with something like

```
mkdir -p $BUNDLE_A/rootfs/some/path
chmod 000 $BUNDLE_A/rootfs/some/path
umoci repack # whatever
umoci insert blah /some/path/blah
```

(This should make no difference to the actual generation, but it would ensure that rootless is "handled" in that case. Also a similar case should be handled from the host side of things to make sure rootless is properly obeyed and the added directory is actually reconstructed correctly.)

Also this test should do an `unpack` after the first insertion to ensure that the file actually changes between the two insertions IMO.

If all of that is a bit too much to change, I can push them on top of your branch once you're done developing.

oci/layer/generate.go
```diff
@@ -97,3 +101,44 @@ func GenerateLayer(path string, deltas []mtree.InodeDelta, opt *MapOptions) (io.
 
 	return reader, nil
 }
+
+// GenerateInsertLayer generates a completely new layer from the root path to
+// be inserted into the image at "prefix".
+func GenerateInsertLayer(root string, prefix string, opt *MapOptions) io.ReadCloser {
+	root = path.Clean(root)
+	rootPrefixLen := len(root) - len(path.Base(root))
+	prefix = strings.TrimPrefix(prefix, "/")
```
cyphar: This should be some mix of `utils.CleanPath` similar. I'm not sure which. I get that you want to strip leading '/' but you can do that in the `Join` call. In addition, I'm not really sure what you're doing with `rootPrefixLen` -- why not use `path.Rel`?

cmd/umoci/insert.go
```diff
@@ -0,0 +1,140 @@
+/*
+ * umoci: Umoci Modifies Open Containers' Images
+ * Copyright (C) 2018 Cisco
```
cyphar: A lot of this file is copied from the other `cmd/umoci/insert.go` files -- can you add an additional copyright line from the original?

cmd/umoci/insert.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	// Create the mutator.
+	mutator, err := mutate.New(engine, descriptorPaths[0])
+	if err != nil {
+		return errors.Wrap(err, "create mutator for base image")
+	}
+
+	// TODO: add some way to specify these from the cli
+	reader := layer.GenerateInsertLayer(ctx.Args()[0], ctx.Args()[1], nil)
```
cyphar: You always pass `nil` for `MapOptions` here. That doesn't look right to me -- I would copy the `unpack` code for map options (perhaps even refactor them out and use the same code). I'm also not a huge fan of the `ctx.Args` usage here -- I would use `ctx.App.Metadata` (as ugly of a hack it might be) to make it more consistent.

cmd/umoci/insert.go
```diff
@@ -0,0 +1,140 @@
...
+	Before: func(ctx *cli.Context) error {
+		if ctx.NArg() != 2 {
+			return errors.Errorf("invalid number of positional arguments: expected <file> and <path>")
+		}
+		if ctx.Args()[0] == "" {
+			return errors.Errorf("path cannot be empty")
+		}
+		if ctx.Args()[1] == "" {
+			return errors.Errorf("path cannot be empty")
+		}
```
cyphar: I would specify which path cannot be empty. Also, I'm not sure why the prefix cannot be empty -- an empty prefix just means it should be added to the root, no?

cmd/umoci/insert.go
```diff
@@ -0,0 +1,140 @@
...
+
+var insertCommand = uxHistory(cli.Command{
+	Name:  "insert",
+	Usage: "insert a file into an OCI image without unpacking/repacking it",
+	ArgsUsage: `--image <image-path>[:<tag>] <file> <path>
+
+Where "<image-path>" is the path to the OCI image, "<tag>" is the name of the
+tag that the content wil be inserted into (if not specified, defaults to
+"latest"), "<file>" is the file or folder to insert, and "<path>" is the prefix
+inside the image to insert it into.`,
```
cyphar: I would mention that the prefix is *not* like `tar` prefixes -- where you can make prefixes that aren't entire path components. I assume this decision was intentional, so I would mention it.

test/insert.bats
tych0: Ok, might be worth changing unpack/repack too, that's where I got this from, and they're just inlined there.

test/insert.bats
tych0: Nope, I'm happy to do that too.

oci/layer/generate.go
tych0: rootPrefixLen is basically some fancy math so that if the user does: `umoci insert --image ... /thing/to/insert /usr/bin`, we end up with `insert` in `/usr/bin`, and not `/usr/bin/thing/to/insert`; `path.Rel` doesn't quite do the right thing here.

The TrimPrefix bit is because of this hunk: https://github.com/openSUSE/umoci/blob/master/oci/layer/tar_generate.go#L94 which causes a failure if there's an absolute path passed to the tar generator. It's not really clear why that's there, but I assume you had a reason, so I just trimmed the prefix. I'm not sure any path cleaning will do that, though.

cmd/umoci/insert.go
tych0: Sure, np

cmd/umoci/insert.go
tych0: I'm not sure what you mean by this, can you elaborate?

cmd/umoci/insert.go
tych0: Not only is it ugly, I think it's also unecessary. You can use https://godoc.org/github.com/urfave/cli#Context.GlobalBool and friends to access global stuff from a local context.

test/insert.bats
cyphar: Oh, then scratch that (I'll double-check why I thought that was the case...).

oci/layer/generate.go
cyphar: > umoci insert --image ... /thing/to/insert /usr/bin, we end up with insert in /usr/bin, and not /usr/bin/thing/to/insert; path.Rel doesn't quite do the right thing here.

Shouldn't be able to just do `filepath.Base`?

> The TrimPrefix bit is because of this hunk: https://github.com/openSUSE/umoci/blob/master/oci/layer/tar_generate.go#L94 which causes a failure if there's an absolute path passed to the tar generator. It's not really clear why that's there, but I assume you had a reason, so I just trimmed the prefix.

It's for security reasons, to double-check that the path lexically doesn't have any un-clean components. But you're right that I didn't consider the `"/"+"/abc" != filepath.Join("/", "/abc")` edge-case (because generally tar archives don't have leading `/` for their entries).

We could fix this by removing leading `/` when normalising the paths.

tych0: no, because `filepath.Base("/thing/to/insert") == "insert"`, not `"/thing/to"`, which is what we want to know the length of. I think `filepath.Dir()` is the right thing though, I can change it to that.

R.e. the tar issue, you do a CleanPath() before that check, so shouldn't all the ..s be cleaned out? I suppose extra paranoia is ok, but if you don't trust your own code...

I'm happy to just trim the leading / in normalise(), though.

cmd/umoci/insert.go
tych0: Ok. What do you want to see, then? I personally detest rsync's whole thing about trailing slash vs not, so I'd like to avoid that if possible.

tych0:
IMO any weirdness about the user specifying a trailing / is basically terrible, and we should never do that. The semantics aren't the same as cp or rsync, but rsync's interface sucks so I'll ignore that. I could imagine something like `umoci insert /path/to/insert /path/to/copy/to`, which would be identical to `cp`, and thus require you to specify whatever prefix you want. FWIW, that's exactly what happens now, it's just that the documentation describes it in terms of prefix instead of destination path.

cyphar:
In short, there are two things that I think should be doable with `umoci insert`. If you like we can work on these separate to this PR or we can work on them now. Here they are:

```
% # inserts a file (or directory) *at* the given path -- not as a child of the directory
% umoci insert --image some_image my_new_etc_passwd /etc/passwd
% # inserts a directory, overwriting any pre-existing directory -- not just merging the new with the old
% umoci insert --image some_image --whiteout some_dir /var/existing_directory
```

The first is just a matter of the path we generate, and making it handle that case. The second requires generating an opaque whiteout for `/var/existing_directory`.

> IMO any weirdness about the user specifying a trailing / is basically terrible, and we should never do that. The semantics aren't the same as cp or rsync, but rsync's interface sucks so I'll ignore that.

FWIW, I agree that `rsync`'s interface is pretty annoying, but I can't think of a simpler way of disambiguating between "I want you to insert this file **at** this path" versus "I want you to insert this file **under** this path". If you have one that doesn't require replicating the lovely `/` confusion, I'd be happy to take it -- I just couldn't think of a better example. :confounded:

But in my view the above two operations are pretty important (though the first is mainly a quality-of-life thing) in order to be able to use this in all of the situations I can think of. But as I said above, we can merge these changes earlier and I can work on the other things before the next release -- whichever works for you.

cyphar:
> FWIW, I agree that `rsync`'s interface is pretty annoying, but I can't think of a simpler way of disambiguating between "I want you to insert this file **at** this path" versus "I want you to insert this file **under** this path". 

Sorry, it should also be noted that we don't have to deal with the most confusing case with `rsync` -- "what if the target path already exists". Since we don't have the ability to (easily and efficiently) figure out if a path exists, that case is entirely ignored. Really then it boils down to

```
% umoci insert --image ABC /some/path /target/path  # /target/path = /same/path
% umoci insert --image ABC /some/path /target/path/ # /target/path/path = /same/path
```

Which is just a different way of making a boolean flag, like this:

```
% umoci insert --image ABC --verbatim /some/path /target/path # /target/path = /same/path
% umoci insert --image ABC            /some/path /target/path # /target/path/path = /same/path
```

Now, `--verbatim` is more explicit and less likely to be triggered by accident. But you could argue it's more cumbersome. But I don't really mind either way to be honest.

tych0:
On Tue, Apr 24, 2018 at 09:32:08AM -0700, Aleksa Sarai wrote:
> > FWIW, I agree that `rsync`'s interface is pretty annoying, but I can't think of a simpler way of disambiguating between "I want you to insert this file **at** this path" versus "I want you to insert this file **under** this path". 
> 
> Sorry, it should also be noted that we don't have to deal with the most confusing case with `rsync` -- "what if the target path already exists". Since we don't have the ability to (easily and efficiently) figure out if a path exists, that case is entirely ignored. Really then it boils down to
> 
> ```
> % umoci insert --image ABC /some/path /target/path  # /target/path = /same/path
> % umoci insert --image ABC /some/path /target/path/ # /target/path/path = /same/path
> ```
> 
> Which is just a different way of making a boolean flag, like this:
> 
> ```
> % umoci insert --image ABC --verbatim /some/path /target/path # /target/path = /same/path
> % umoci insert --image ABC            /some/path /target/path # /target/path/path = /same/path
> ```
> 
> Now, `--verbatim` is more explicit and less likely to be triggered by accident. But you could argue it's more cumbersome. But I don't really mind either way to be honest.

I guess my argument would be that you basically never want
/target/path/path, and if you do, you can just say that instead. So it
should always be:

target := path.Join(targetPath, path.Base(sourcePath))

of course, it's slightly more complicated, since when someone gives a
directory, we'll end up walking the path, and inside the walk we'll do
something like:

target := path.Join(targetPath, current[path.Dir(sourcePath):])

That's basically what it does now, except with slightly uglier math on
the paths.

Tycho

cyphar:
This will now need a rebase on top of #240, to use the refactored `--uid-map` parsing code (don't worry, it makes it easier).

tych0:
Thanks for the heads up. Also, I had another thought about the `/target/path/path` issue, which was that in fact the current version doesn't. Your contrived example repeats path, but in the real world you would not:

`umoci insert myexecutable /usr/bin`
`umoci insert myconfig /etc`

Whereas with what you're proposing, we need to do,

`umoci insert myexecutable /usr/bin/myexecutable`
`umoci insert myconfig /etc/myconfig`

Again I'm happy to switch it, but it does result in more typing for everyone, although it also allows you to rename things on the fly.

tych0:
Ok, I think I've addressed everything; let me know if I missed something.

tych0:
Ah yes, the best kind of bug, where it passes locally and fails on travis :)

cyphar:
I'll review this later today, or I can wait until you've fixed up the Travis failure (it's failing during rootless tests with bad argument errors, which is a bit odd).

> Again I'm happy to switch it, but it does result in more typing for everyone, although it also allows you to rename things on the fly.

That's why I suggested having both options -- similar to how `cp` works (though with `cp` you can `stat` the target path to figure out if it's a directory or not -- doing that in an image would make it much more expensive). Personally I think it's necessary to have the ability to "rename things on the fly" (we could add a `-T` or `-t` option that matches the `cp` semantics if we didn't want to do it using the trailing-`/` semantics I proposed earlier).

tych0:
On Mon, Apr 30, 2018 at 04:02:48PM -0700, Aleksa Sarai wrote:
> I'll review this later today, or I can wait until you've fixed up the Travis failure (it's failing during rootless tests with bad argument errors, which is a bit odd).

yeah, I don't know what that's about. it works for me locally.

> > Again I'm happy to switch it, but it does result in more typing for everyone, although it also allows you to rename things on the fly.
> 
> That's why I suggested having both options -- similar to how `cp` works (though with `cp` you can `stat` the target path to figure out if it's a directory or not -- doing that in an image would make it much more expensive). Personally I think it's necessary to have the ability to "rename things on the fly" (we could add a `-T` or `-t` option that matches the `cp` semantics if we didn't want to do it using the trailing-`/` semantics I proposed earlier).

yeah, check out the way it works now. it's cp without the stat,
exactly as in one of your proposals.

cyphar:
Awesome. I'll play around with it today, and I can try to figure out why the tests are failing as well.

tych0:
Ok, I think I figured out what was wrong with the tests. This should be good to go now.

tych0:
Ping, any thoughts on this?

cyphar:
This is looking pretty good. I'm going to play around with this and do a proper review soon. This will need a rebase. Documentation would also be good, but if you prefer I can do that in a follow-up.

tych0:
I rebased it. Let's get a version in before we bikeshed on any docs.

cyphar:
Sure. :wink:

cyphar:
One thing that I mentioned above is that we don't have a way of overwriting an entire directory -- so if you insert a directory over an existing directory you end up with a merge of the two (with no nice way of only getting the new directory).

```
% mkdir dir1 && touch dir1/should_not_exist
% mkdir dir2 && touch dir2/should_exist
% umoci insert --image image dir1 /path
% umoci insert --image image dir2 /path
% umoci unpack --image image bundle && ls bundle/rootfs/path
should_exist  should_not_exist
```

I will work on this feature once this is merged as well. I will probably just make it something like a `--whiteout` flag.

cyphar:
Rootless tests are timing out for some reason. Looking into it...

cyphar:
Tests are timing out on Travis (but only for rootless and only for Fedora). I can reproduce the test taking a long time (though it doesn't block indefinitely). Not sure why the test is taking longer than it used to..

cyphar:
Okay, there's now a test failure in rootless (you can't see it on Travis for some reason):

```
not ok 60 umoci repack [replace]
# (in test file test/repack.bats, line 204)
#   `ln -s "a \\really //weird _00..:=path " "$BUNDLE_A/rootfs/usr/bin/env"' failed
# 21M   /tmp/umoci-coverage.7NhRrl
# 0     /tmp/umoci.YCR12A
# 83M   /tmp/image/blobs
# 83M   /tmp/image
# 103M  /tmp
# oci-image-tool: reference "latest": OK
# /tmp/image: OK
# Validation succeeded
# umoci -test.coverprofile=/tmp/umoci-coverage.7NhRrl/umoci.cov.vtQiqE __DEVEL--i-heard-you-like-tests unpack --rootless --image /tmp/image:latest /tmp/umoci-integration-tmpdir.8q2KHYm5 (status=0)
# PASS
# coverage: 31.9% of statements in ./...
# Bundle validation succeeded.
# rm: remove write-protected regular file '/tmp/umoci-integration-tmpdir.8q2KHYm5/rootfs/usr/bin/env'? ln: failed to create symbolic link '/tmp/umoci-integration-tmpdir.8q2KHYm5/rootfs/usr/bin/env': File exists
```

tych0:
Ok, this seems unrelated -- presumably /usr/bin/env is not +w, and since you're not passing -f to the rm of it, it's blocking. Why it works on other platforms I have no idea :)

FWIW, I also see failures on current master on ubuntu of the same `umoci repack` test,
```
# /tmp/bats.8942.src: line 165: /tmp/umoci-integration-tmpdir.4dWrFtb7/rootfs/whiteout_test/.wh. THIS IS A TEST : Operation not permitted
```
I've been ignoring these as well.

cyphar:
#243 fixed the test failures. If you rebase I'll merge it, thanks!

tych0:
Cool, thanks. Looks like the tests passed this time.

cyphar:
LGTM. Thanks @tych0!

PR #239, opened 2018-04-23
umoci: add "raw unpack"

### LINKED ISSUE #23: umoci: implement raw-unpack
This is like `unpack` except it just does a simple extraction of a given layer diff blob (no `mtree` or `config.json`). This probably will end up being in a set of `raw-*` subcommands that are more hardcore versions of the nice interfaces.

flx42:
This is pretty basic, and duplicates quite some code with the existing `unpack` code, that's why I've tagged this as WIP, to get the ball going and discuss our options.

I would like this feature for cases where we only care about getting the final rootfs. Computation of the mtree file can take a significant amount of time on large images (because of all the checksums that need to be computed).

thoughts @cyphar?

Closes #23

### DIFF
--- a/cmd/umoci/raw-unpack.go
+++ b/cmd/umoci/raw-unpack.go
@@ -1,0 +1,131 @@
+/*
+ * umoci: Umoci Modifies Open Containers' Images
+ * Copyright (C) 2018 SUSE LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/apex/log"
+	"github.com/openSUSE/umoci/oci/cas/dir"
+	"github.com/openSUSE/umoci/oci/casext"
+	"github.com/openSUSE/umoci/oci/layer"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
+	"golang.org/x/net/context"
+)
+
+var rawUnpackCommand = uxRemap(cli.Command{
+	Name:  "unpack",
+	Usage: "unpacks a reference into a rootfs",
+	ArgsUsage: `--image <image-path>[:<tag>] <rootfs>
+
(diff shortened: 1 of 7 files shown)


cmd/umoci/raw-unpack.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		meta.MapOptions.UIDMappings = append(meta.MapOptions.UIDMappings, idMap)
+	}
+	for _, gidmap := range ctx.StringSlice("gid-map") {
+		idMap, err := idtools.ParseMapping(gidmap)
+		if err != nil {
+			return errors.Wrapf(err, "failure parsing --gid-map %s: %s", gidmap)
+		}
+		meta.MapOptions.GIDMappings = append(meta.MapOptions.GIDMappings, idMap)
+	}
```
cyphar: Since this code is now used by `raw unpack` (and will be used by `insert` in #237) this should be moved to a separate function.

cyphar:
I'm totally fine with having a new `raw unpack` command -- as you said it's something that quite a few people have asked for. The current status looks fine (though there's a bit of duplication we should fix). Not to mention this obviously needs tests (you could probably just copy the `unpack` tests and turn them into `raw unpack` tests -- maybe there's a way to refactor it) and documentation.

oci/layer/unpack.go
```diff
@@ -120,19 +117,50 @@ func UnpackManifest(ctx context.Context, engine cas.Engine, bundle string, manif
 		return errors.Wrap(err, "bundle path empty")
 	}
 
+	defer func() {
+		if err != nil {
+			fsEval := fseval.DefaultFsEval
```
cyphar: AFAICS this logic is wrong (if you intended to not use `unpriv.RemoveAll` you could just do `os.RemoveAll`). You need to do something like in `UnpackRootfs`:

flx42: You're right, I think I was misunderstanding why this particular is cleanup is needed.
Are you ok with a little duplication between the two function, or should we put that in a helper function `removeAll`?

cyphar: I'd just duplicate it, personally.

flx42:
Ok, so fixed the deferred removal. Added a few tests. But it depends on #240 
Next would be writing the doc, if we agree on the rest.

flx42:
Rebased on top of current `master`, added the man page.
Do we need more tests?

Also, it keeps the old behavior of failing if the target rootfs folder already exists, are you OK with that?

cyphar:
This needs a rebase (again) -- sorry about that. The tests look fine to me.

> Also, it keeps the old behavior of failing if the target rootfs folder already exists, are you OK with that?

Yes I'm okay with that. We can always change it later.

cyphar:
LGTM. Will merge after test success.

flx42:
Sorry, with DockerCon last week I didn't have time to do the rebase.

cyphar:
No problem. It was trivial to rebaseand it's all been merged. I will do a release with these features once I clean up the `umoci insert` stuff (it **needs** opaque whiteout support, and documentation, as well as a little bit of refactoring).

PR #240, opened 2018-04-25
Refactor ID mapping

flx42:
Preliminary work for #239

### DIFF
--- a/cmd/umoci/raw-runtime-config.go
+++ b/cmd/umoci/raw-runtime-config.go
@@ -25,14 +25,13 @@ import (
 	"github.com/openSUSE/umoci/oci/cas/dir"
 	"github.com/openSUSE/umoci/oci/casext"
 	"github.com/openSUSE/umoci/oci/layer"
-	"github.com/openSUSE/umoci/pkg/idtools"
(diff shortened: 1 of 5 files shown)


cyphar:
Please run `gofmt` on the changed files -- that's what's causing the CI to fail. It would also be nice if we could put the new code inside `utils_ux.go` (and possibly making it follow the existing `uxImage` style where it is automatically applied based on the command type, thought I don't think we can abuse the categories for this...).

cyphar:
> Should we also move the ID parsing function to utils_ux.go though? This doesn't seem really right for me.

No, we could put it into `utils.go`. However, *maybe* we should use the same `ctx.App.Metadata` trick that we use for other parsed options. The downside is that you'd need to stuff in a `MapOptions` struct into it. This is something I could look into refactoring later if it makes it easier for you.

> I don't think we can reuse the category mechanism for this since it can't be chained in its current form, the decorator approach can.

I will try to figure out a work-around for this after we merge it as well (since it's more of a maintenance thing than anything else).

cyphar:
Sure, that's fine (though for reference the Metadata map is of `interface{}`, not `string`). Will review in a bit.

cyphar:
LGTM.

PR #243, opened 2018-05-15
test: force removal to fix Travis CI failure

cyphar:
Apparenty Fedora has made some of their binaries unwriteable for all
users, which results in the rootless tests failing with strange errors:

    not ok 60 umoci repack [replace]
    # rm: remove write-protected regular file '/tmp/umoci-integration-tmpdir.8q2KHYm5/rootfs/usr/bin/env'?
    # ln: failed to create symbolic link '/tmp/umoci-integration-tmpdir.8q2KHYm5/rootfs/usr/bin/env': File exists

Fix this by just forcing removal everywhere in the tests.

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -159,7 +159,7 @@ function teardown_tmpdirs() {
 	done < "$TESTDIR_LIST"
 
 	# Clear tmpdir list.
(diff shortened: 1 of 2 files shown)


cyphar:
LGTM.

PR #245, opened 2018-06-26
add --keep-dirlinks option to unpack

tych0:
Similar to rsync's --keep-dirlinks, let's have an option to allow symlinks
that point to directories to be used as directories themselves.


tych0:
Ping, just occurred to me that I hadn't heard anything on this.

cmd/umoci/unpack.go
```diff
@@ -49,7 +49,12 @@ creation with umoci-repack(1).`,
 	// unpack reads manifest information.
 	Category: "image",
 
-	Flags: []cli.Flag{},
+	Flags: []cli.Flag{
+		cli.BoolFlag{
+			Name:  "keep-dirlinks",
```
cyphar: This flag should probably also be added to `raw-unpack`. And docs should be updated (it doesn't need much).

oci/layer/tar_extract.go
```diff
@@ -370,7 +396,13 @@ func (te *tarExtractor) unpackEntry(root string, hdr *tar.Header, r io.Reader) (
 	// "lower" file still exists (so the hard-link would point to the old
 	// inode). It's not clear if such an archive is actually valid though.
 	if !fi.IsDir() || hdr.Typeflag != tar.TypeDir {
-		if err := te.fsEval.RemoveAll(path); err != nil {
+		if fi.Mode()&os.ModeSymlink != 0 && hdr.Typeflag == tar.TypeDir &&
+			te.mapOptions.KeepDirlinks {
```
cyphar: To me, this nested if looks a bit complicated (and is what stopped me from completing my review when you first sent the PR -- because I tried to come up with a more "clear" conditional and didn't get very far).

At the moment, I'm +1 on the design it just needs to be a bit less difficult to understand. I'll take another look and get back to you.

cmd/umoci/unpack.go
tych0: Cool, I pushed a commit for these.

oci/layer/tar_extract.go
tych0: Sure, no worries. Perhaps a comment explaining that we want to write through symlinks instead of overwriting them is enough?

cmd/umoci/unpack.go
cyphar: Thanks. :D

oci/layer/tar_extract.go
tych0: Ok, I added the comment (back from vacation now). If you merge this, I have the v2 api done on top of this and I can post that too.

tych0:
Ping, any thoughts on this?

tych0:
Ping, any thoughts on this?

cyphar:
Needs a rebase, but I can do that. I don't really like the re-use of `MapOptions` for something unrelated to mapping -- but that just requires a struct type rename. I can fix that up later.

tych0:
Ok, I've rebased it. It turns out that renaming MapOptions isn't so easy, because the struct is emitted in UmociMeta, so we'd have to write some backwards compatible code for that. For now I just disabled it being emitted in the json, since it's not really relevant, and I'll let you decide how to handle it in the future as you say.

oci/layer/tar_extract.go
```diff
@@ -370,7 +396,19 @@ func (te *tarExtractor) unpackEntry(root string, hdr *tar.Header, r io.Reader) (
...
+		// is a symlink, just write through instead of removing it and
+		// creating a directory. (The idea being that e.g. sometimes
+		// /lib64 is a symlink, and sometimes it's a real directory,
+		// but in both cases you want all the libraries, and not just
+		// the ones that were under a real directory.)
+		if fi.Mode()&os.ModeSymlink != 0 && hdr.Typeflag == tar.TypeDir &&
+			te.mapOptions.KeepDirlinks {
+			isDirlink, err = te.isDirlink(root, path, hdr)
+			if err != nil {
+				return err
```
cyphar: Needs an `errors.Wrap(err, "check is dirlink")`.

oci/layer/tar_extract.go
```diff
@@ -167,6 +167,31 @@ func (te *tarExtractor) applyMetadata(path string, hdr *tar.Header) error {
...
+	if err != nil {
+		return false, errors.Wrap(err, "read existing link")
+	}
+
+	linkPath, err := securejoin.SecureJoinVFS(root, link, te.fsEval)
+	if err != nil {
+		return false, errors.Wrap(err, "sanitize old target")
+	}
+
+	linkInfo, err := te.fsEval.Lstat(linkPath)
```
cyphar: I just realised that this won't handle nested symlinks (a symlink to a symlink to a directory) properly -- we'd need to loop this logic and bail with an `ELOOP`. (I really wish we had `openat(..., AT_THIS_ROOT)` so we didn't have to write this logic in userspace.)

test/unpack.bats
```diff
@@ -152,3 +152,35 @@ function teardown() {
 
 	image-verify "${IMAGE}"
 }
+
+@test "umoci unpack --keep-dirlinks" {
```
cyphar: Once the nested symlink thing is fixed, I'd recommend adding a test for it.

tych0:
Ok, my rebuild of this worked without http flakiness. I don't have perms to retrigger the travis build, but I think if you do it'll work and should be mergable.

cyphar:
LGTM.

cyphar:
LGTM.

cyphar:
I was just re-reading the tests and it looks like they actually don't quite test `--keep-dirlink`. But don't worry, I'll work on this (it'll require having an `umoci raw add-layer` or something where we add a tar archive generated externally -- because `umoci repack` won't generate a dirlink and so testing it with just `umoci repack` won't quite work).

PR #246, opened 2018-08-02
Api v2

tych0:
Hi,

This is a (much smaller and less fiddly) alternative to #219. The basic idea is that we don't wrap anything: we just expose `casext.Engine` and let people muck with that, the same way umoci itself does.

I've also included some public-ifying of additional code that I found useful.

Thoughts are welcome, but hopefully this is much less obnoxious than the other one :)

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,0 +1,30 @@
+package umoci
+
+import (
+	"github.com/openSUSE/umoci/oci/cas/dir"
+	"github.com/openSUSE/umoci/oci/casext"
+	"github.com/pkg/errors"
+)
+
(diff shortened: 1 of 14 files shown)


tych0:
I rebased this after the dirlinks merge :)

cyphar:
There was a minor CI failure, I've fixed it.

cyphar:
@tych0 We should probably add a copyright line to the new files -- what line would you find appropriate?

tych0:
Copyright (C) Cisco Systems 2018

Is good :-)

Thanks!

PR #247, opened 2018-08-02
vendor: un-fork everything

### LINKED ISSUE #244: Needless fork of github.com/pkg/errors ?
Vendored [github.com/pkg/errors](https://github.com/pkg/errors) appears to be a fork since building against pristine library produces the following error:

~~~~
# github.com/openSUSE/umoci/cmd/umoci
src/github.com/openSUSE/umoci/cmd/umoci/main.go:97:4: undefined: "github.com/pkg/errors".Debug
~~~~

Indeed vendored library appears to be modified as I could not find traces of `Debug` in the original library repository.

Needless forking is harmful. Please consider using unmodified library.

If you have to (use the) fork of the original library for whatever reason then at least please don't hijack its import path.

Thank you.

cyphar:
Since the start of this project, we've been using a pretty dodgy vendoring
script I wrote to work around issues with `glide` and other such tools that we
used in the past. Since then, tools like `vndr` have become more popular and
it's come time to switch.

In addition, we were carring a few patches of upstream projects that could be
considered a bad maintainence practice -- so we've dropped them (even though
one of the patches makes the binary footprint of umoci much smaller due to a
massively reduced dependency tree).

Closes #244.


cyphar:
LGTM.

PR #252, opened 2018-08-29
oci: ignore system.nfs4_acl and extend forbidden-xattr handling

### LINKED ISSUE #248: Unpack on NFS and Extended File Attributes
Hi all,

## Issue

I tried to use umoci to extract images on an NFS file-system and encountered various errors linked to extended file attributes. Umoci fails to unpack in both rootless and default cases.

## Reproducer

You need to be on an NFS mount.

```bash
#Get alpine
$ skopeo copy docker://alpine oci:alpine_image:latest
#!
#! Unpack in rootless
#!
$ umoci --log debug  unpack --rootless --image ./alpine_image ./alpine_bundle           1 ↵
   • parsed mappings           map.gid=[{1000 0 1}] map.uid=[{1000 0 1}]
   • -> ws.recurse             digest=sha256:8327fe6785fae2e722377350939dc07cd83674d285f537e1a91893248b925bc3
   • <- ws.recurse             digest=sha256:8327fe6785fae2e722377350939dc07cd83674d285f537e1a91893248b925bc3
   • casext.ResolveReference(latest) got these descriptors refs=[{[{application/vnd.oci.image.manifest.v1+json sha256:8327fe6785fae2e722377350939dc07cd83674d285f537e1a91893248b925bc3 348 [] map[org.opencontainers.image.ref.name:latest] 0xc420088900}]}]
   • umoci: unpacking OCI image bundle=./alpine_bundle image=./alpine_image ref=latest rootfs=rootfs
   • unpacking bundle ...     
   • unpack rootfs: alpine_bundle/rootfs
   • unpack layer: sha256:8e3ba11ec2a2b39ab372c60c16b421536e50e5ce64a0bc81765c2e38381bcff6
   • unpacking entry           path=bin root=alpine_bundle/rootfs type=53
   ⨯ create runtime bundle: unpack rootfs: unpack layer: unpack entry: bin: apply hdr metadata: clear xattr metadata: alpine_bundle/rootfs/bin: unpriv.lclearxattrs: unpriv.lremovexattr: input/output error
#!
#! Unpack as root
#!
(as root) umoci --log debug  unpack --image ./alpine_image ./alpine_bundle
[...] #Same error
   ⨯ create runtime bundle: unpack rootfs: unpack layer: unpack entry: bin: apply hdr metadata: clear xattr metadata: alpine_bundle/rootfs/bin: unpriv.lclearxattrs: unpriv.lremovexattr: input/output error
```
## Discussion

This is linked to the fact that NFS does not support extended file attributes. When removing 'system.nfs4_acl' in Lclearxattrs you get _EIO_ and overall when setting any xattr in Lsetxattr you get _ENOTSUP_.

## Proposed Fix

I humbly tempted an initial fix which seems to solve the issue on my system in this branch:

https://github.com/besnardjb/umoci/tree/nfs

- *Fseval Default:* https://github.com/besnardjb/umoci/commit/3ca29fdccbfba4a9b19529eebe68648002aa48f5
- *Fseval Rootless:* https://github.com/besnardjb/umoci/commit/612c1d229580c7df771f63f4970eeb9c6de0b457

However, as it alters the resulting image (i.e. losing xattr data) it might deserve a warning as some data of the image was lost even if not representable in the target fs?

Thanks,

Jean-Baptiste.

cyphar:
It turns out that system.nfs4_acl is yet another xattr that we shouldn't
be touching, but also that we should not be clearing (while we have
permission to do so, clearing it results in NFS permissions breaking).
So as an extension we now now use ignoreXattrs for both unpack and
repack operations.

Closes #248


cyphar:
Alright, this should fix the issue. @besnardjb did you want to test this?

besnardjb:
Hi @cyphar,

Thank you very much for the pull request, it is indeed much cleaner.

I've made a quick test on the Centos OCI image (from docker) and the NFS part is solved for me when extracting on shared FS with your configurable filter.

However, I've encountered another attribute (user.rootlesscontainers) in this same image. As a consequence, I still fail on *lsetxattr* as attributes are not supported on the target file-system (details below). 

Here is how it failed and how I got there:

```
$ mount | grep raid
(...):/media/raid on /media/raid type nfs4 (rw,relatime,vers=4.2,rsize=1048576,wsize=1048576,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=10.19.213.42,local_lock=none,addr=10.19.213.127)
$ pwd
/media/raid/phomes/jbbesnard/(...)/umoci
$ git describe                                                                                           
v0.4.1-19-gaa8200d
$ skopeo copy docker://centos oci:centos:latest
(...)
$ LANG=us strace -f -e trace=lsetxattr ./umoci --log debug  unpack --rootless --image ./centos ./centos_bundle
(...)
   • rootless{run/} ignoring forbidden xattr: "system.nfs4_acl"
   • unpacking entry           path=run/lock root=centos_bundle/rootfs type=53
   • rootless{run/} ignoring forbidden xattr: "system.nfs4_acl"
   • unpacking entry           path=run/lock/lockdev root=centos_bundle/rootfs type=53
[pid  4563] lsetxattr("centos_bundle/rootfs/run/lock/lockdev", "user.rootlesscontainers", "\10\377\377\377\377\17\0206", 8, 0) = -1 EOPNOTSUPP (Operation not supported)
   • rootless{lock/} ignoring forbidden xattr: "system.nfs4_acl"
   ⨯ create runtime bundle: unpack rootfs: unpack layer: unpack entry: run/lock/lockdev: apply hdr metadata: restore xattr metadata: centos_bundle/rootfs/run/lock/lockdev: unpriv.lsetxattr: operation not supported
(...)
+++ exited with 1 +++
#Same on alpine:
[pid  4956] lsetxattr("alpine_bundle/rootfs/etc/shadow", "user.rootlesscontainers", "\10\377\377\377\377\17\20*", 8, 0) = -1 EOPNOTSUPP (Operation not supported)

```

As my FS returns EOPNOTSUPP on such calls, I fail on extracting the image. In fact, it seems that exrtended attributes are not supported on NFS v4. 

So (and to my knowledge) I see some ways here:
- I _manually_ add "user.rootlesscontainers" to the ban list as I'm on NFS but it would surely be useful on systems supporting it;
- Add a flag basically telling "do your best I WANT this rootfs at all costs" 🤣 -- permissive;
- As the target FS says that it does not support xattrs (EOPNOTSUPP) umoci emits a warning that information is being lost and just keep going on.

The later was my strategy here (https://github.com/besnardjb/umoci/commit/612c1d229580c7df771f63f4970eeb9c6de0b457) although the implementation is clearly disputable 😄 !

```
// On NFS altering extended attrs may not be supported
if errors.Cause(err).Error() == "operation not supported" {
    continue
}
```

Thanks for your great work !

I remain available.

Cheers.

cyphar:
Ah okay. Yeah, we should ignore (but print a warning) on ENOTSUP. The string comparison is not the best way of doing it -- but I do understand what you mean. I'll work on an updated patch for that.

cyphar:
Okay, I've pushed a new patch that should fix it.

besnardjb:
I confirm this solved my issue on both images for an NFS extract. Thanks!

cyphar:
LGTM.

PR #255, opened 2018-09-04
cmd: add 'umoci raw add-layer'

cyphar:
This is a very simple helper function that allows users to "BYO" layers
that were generated elsewhere. It should be noted that this does
slightly hardcode tar archives into umoci's interface more than I'd
like, but I'm sure we can work around that if we have to in the future.

In addition, this fixes the --keep-dirlinks tests to test all of the
important features (like nesting resolution as well as loop detection
and breakout).

/cc @tych0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,8 @@ RUN zypper -n in \
 		oci-image-tools \
 		oci-runtime-tools \
 		python-setuptools python-xattr attr \
-		skopeo
+		skopeo \
+		tar
 
 ENV GOPATH /go
 ENV PATH $GOPATH/bin:$PATH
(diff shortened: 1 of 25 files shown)


cyphar:
@tych0 It looks like my suspicion was correct -- the `--keep-dirlinks` code doesn't handle symlink loops quite right (to be fair, my example test is pretty mean but it does demonstrate the issue). I'll work on a fix for it -- just thought you aught to know since you're the only user of this crazy feature. :wink:

cyphar:
@tych0 would you mind taking a look at "unpack: fix symlink loop handling in --keep-dirlinks" to see whether that patch breaks anything you're using and to see if it makes sense to you? Thanks.

cyphar:
LGTM.

cyphar:
LGTM. I'm going to merge this. @tych0 if you have any complaints about this feel free to create an issue.

tych0:
This does seem to break things for me.

PR #258, opened 2018-09-09
*: various whiteout improvements

### LINKED ISSUE #254: Bad whiteout when importing in docker
The umoci version in question is 0.4.1
Steps to reproduce:
```
umoci init --layout test
umoci new --image test:latest

sudo umoci unpack --image test:latest test-bundle1
sudo mkdir test-bundle1/rootfs/test
sudo touch test-bundle1/rootfs/test/test
sudo umoci repack --image test:latest test-bundle1

sudo umoci unpack --image test:latest test-bundle2
sudo rm -rf test-bundle2/rootfs/test
sudo umoci repack --image test:latest test-bundle2

sudo skopeo --insecure-policy copy oci:test:latest docker-archive:test.tar
cat test.tar | docker load
```

Docker will output the following error: "Error processing tar file(exit status 1): not a directory"
If I tar tvf the second layer, I get:
drwxr-xr-x root/root         0 2018-09-03 16:25 .
---------- 0/0               0 2018-09-03 16:25 .wh.test
---------- 0/0               0 2018-09-03 16:25 test/.wh.test
I believe test/.wh.test should not be present

### LINKED ISSUE #257: umoci-insert needs to support (opaque) whiteouts
Basically it should be possible to do both of these:

```
% umoci insert --opaque ./some/dir /etc # mask old /etc/* entries
% umoci insert --whiteout /opt/blah     # delete this entry
```

The former should mask away all of the previous entries under /etc (by adding an `/etc/.wh..opq` to the layer), while the latter should create a whiteout of the path `/opt/blah`. This is necessary for a generally-usable insertion system.

Also we should consider how this will interact with the long-term goal of #256.

cyphar:
Previously umoci-insert(1) was quite limited in that it would only
insert a given file or directory without having any configurability.
Something users might want is the ability to delete a path without an
unpack-repack cycle. Or they want to replace the contents of a directory
rather than merge them.

To this effect, add --whiteout and --opaque which provide the
corresponding features. --whiteout just adds a whiteout at the target
path, while --opaque adds an opaque whiteout for the target before doing
the copy.

In addition, add support for "simplified whiteouts". This change ultimately
makes our images a little bit smaller by no longer whiting-out every sub-entry
when a directory has been deleted.  While this is a net benefit, the cause of
this change is a Docker breakage (which arguably added a restriction that is
contrary to the OCI specification and thus it is definitely *their* bug). But
to play nice (and to take the opportunity to improve some images) we can fix it
in umoci.

Fixes #257 
Fixes #254 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `umoci unpack --keep-dirlinks` (in the same vein as rsync's flag with
   the same name) which allows layers that contain entries which have a symlink
   as a path component. openSUSE/umoci#246
+- `umoci insert` now supports whiteouts in two significant ways. You can use
+  `--whiteout` to "insert" a deletion of a given path, while you can use
+  `--opaque` to replace a directory by adding an opaque whiteout (the default
+  behaviour causes the old and new directories to be merged).
+  openSUSE/umoci#257
 
 ## Fixed
 - Docker has changed how they handle whiteouts for non-existent files. The
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   the same name) which allows layers that contain entries which have a symlink
   as a path component. openSUSE/umoci#246
 
+## Fixed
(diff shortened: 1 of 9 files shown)


cyphar:
LGTM. I've tested #254 and it no longer happens with these patches.

PR #266, opened 2018-09-24
umoci: allow ':' in --image (and similar flags)

### LINKED ISSUE #265: Incompatibilities with colons (and possibly other characters) in extended scope reference names
#234 expanded the scope of possible reference names, however there are several commands/parameters that still do not fully support them. I am trying to use references of the form `test.com/proj:1.2.3` and the colon is causing problems. They work fine as the second parameter to the tag command, but not when copying a tag with a colon or creating a new tag with the new command.

```
$ umoci -v
umoci version 0.4.2
$
```

Works:
```
$ umoci tag --image blah:latest asd/qwe:ew
$
```
Doesn't work:
```
$ umoci tag --image blah:asd/qwe:ew testing
...
   ⨯ invalid --image: path contains ':' character: 'blah:asd/qwe'
$ umoci new --image blah:asd/qwe:ew
...
   ⨯ invalid --image: path contains ':' character: 'blah:asd/qwe'
$

cyphar:
9d126685960c ("oci: casext: extend scope of valid reference names")
added support for having ':' in reference names (in accordance with the
OCI specification updates), but unfortunately it missed that the primary
flag that interacts with reference names (--image) didn't support having
colons in the image tag.

Fixes #265

### DIFF
--- a/cmd/umoci/utils_ux.go
+++ b/cmd/umoci/utils_ux.go
@@ -140,7 +140,7 @@ func uxImage(cmd cli.Command) cli.Command {
 			image := ctx.String("image")
 
 			var dir, tag string
-			sep := strings.LastIndex(image, ":")
+			sep := strings.Index(image, ":")
 			if sep == -1 {
 				dir = image
 				tag = "latest"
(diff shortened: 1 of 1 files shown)


cyphar:
LGTM.

PR #270, opened 2018-11-07
*: add --no-history to disable history auto-appending

cyphar:
Some users (like KIWI) want to end up with an image that only has a
single history entry for the one layer they have generated, but
umoci-config(1) normally generates a config entry for each invocation.

To allow for this usecase, a new --no-history flag was added which just
avoids adding the history in mutate/. Note that a warning will be
printed when using this flag with commands that add new layers because
layers-without-history-entries can cause lots of problems when looking
at the image information (especially if a later image is added with a
history entry -- the history entry will be incorrectly attributed to the
earlier layer).


cyphar:
LGTM.

PR #271, opened 2018-11-07
UnpackRootfs: allow unpacking over existing rootfs

tych0:
We have cases where we want to unpack layers over the top of some stuff
that we've already put on the disk. Add a flag to allow this.

### DIFF
--- a/oci/layer/unpack.go
+++ b/oci/layer/unpack.go
@@ -129,6 +129,13 @@ func UnpackManifest(ctx context.Context, engine cas.Engine, bundle string, manif
 		}
 	}()
 
+	if _, err := os.Lstat(rootfsPath); !os.IsNotExist(err) {
+		if err == nil {
+			err = fmt.Errorf("%s already exists", rootfsPath)
+		}
+		return err
+	}
+
 	log.Infof("unpack rootfs: %s", rootfsPath)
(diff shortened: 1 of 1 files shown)


oci/layer/unpack.go
```diff
@@ -149,18 +149,19 @@ func UnpackManifest(ctx context.Context, engine cas.Engine, bundle string, manif
 
 // UnpackRootfs extracts all of the layers in the given manifest.
 // Some verification is done during image extraction.
-func UnpackRootfs(ctx context.Context, engine cas.Engine, rootfsPath string, manifest ispec.Manifest, opt *MapOptions) (err error) {
+func UnpackRootfs(ctx context.Context, engine cas.Engine, rootfsPath string, manifest ispec.Manifest, opt *MapOptions, allowExistingRoot bool) (err error) {
```
cyphar: I'd prefer if this wasn't done through a boolean flag. Since you're the only user of `UnpackRootfs` I'm fine with just always allowing unpacking into an existing rootfs, and inside `UnpackManifest` we can do the `os.Lstat(...) == nil` check because there we are operating with bundles.

tych0: fixed, thanks.

cyphar:
LGTM. I was going to suggest an explicit check that it's a directory, but that's not really necessary. Thanks! I'm going to make a release in a bit.

PR #272, opened 2018-11-11
*: go1.11 update and module activation

### LINKED ISSUE #269: go1.11 breaks mutate_test.go
I had a feeling this would eventually happen. In `go1.11`, `archive/tar` now generates slightly different archives with our test-cases an as a result our hard-coded hashes no longer pass tests. This is what is stopping Fedora 29 (and rawhide) builds from working...

/cc @ganto

cyphar:
In Go 1.11, the semantics of archive/tar [were changed in a very worrying
way][1]. hdr.Typeflag values of TypeRegA are now silently converted to
TypeReg, which results in changes in the hashes of said objects. Luckily
this didn't affect the core oci/ code because we explicitly used TypeReg
-- but it did affect one of our tests.

This is another justification for pushing through with OCIv2, because
depending on a format that has no guaranteed canonical format is quite
dangerous. I do find it somewhat silly that a change like this was made
for such a minor reason (to better conform with one sentence of the PAX
specification -- in a format where nobody follows the rules). But I do
understand why they'd want to do that.

We also now switch to using Go 1.11 modules. We will still include the
vendor/ directory for the forseeable future (at least until it is sanely possible
to switch away from it entirely) but it's currently still needed for offline builds.

Hopefully I don't regret this. :P

[1]: https://go-review.googlesource.com/c/go/+/85656

Fixes #269


cyphar:
LGTM.

PR #273, opened 2018-11-11
umoci: add --tag to umoci-insert

cyphar:
Before this there was no way to make --tag non-destructive, so we just
copy the semantics from umoci-config. This won't break existing scripts
because the default is still to override (just like it is with
umoci-config).


cyphar:
LGTM.

cyphar:
LGTM.

PR #279, opened 2018-12-16
Update gomodule checksums for go 1.11.4

CameronNemo:
Without this patch, umoci fails to build from source with go 1.11.4.

go 1.11.4 fixes a bug where the checksums for certain types of go modules are incorrectly generated. As such, the go.sum file must be regenerated.

See the following for more information:

https://github.com/golang/go/issues/29278

https://github.com/golang/go/issues/29282


cyphar:
LGTM. Probably a silly question, but it will build okay on older versions of `go` right? It was just a generation bug?

cyphar:
LGTM.

CameronNemo:
Okay, I have run `go mod vendor`.

cyphar:
LGTM, thanks!

PR #282, opened 2018-12-27
oci: cas*: add blob verification

### LINKED ISSUE #278: security: expand CAS blob verification
Currently we verify that layer hashes match (against the DiffID), but it's probably a good idea that we also verify that the CAS blobs all are correct. We can probably do this by wrapping the returned blob with a verifier that returns an error on EOF if the digest is mismatched (though we'd have to make sure our `Close` calls as well as any `io.Copy` errors are all correctly checked).

cyphar:
Previously we didn't really check that blobs matched their hashes
directly -- most of the checks we did were based on layer DiffIDs and
similar checks.

Here we add a new casext.GetVerifiedBlob() API which wraps the
underlying reader with a VerifiedReadCloser that will verify that the
digest matches on EOF (or on Close). Users should be very careful to
ensure that they actually check all errors (such as when using
ioutil.Discard or when they do Close).

In addition, we add it to the underlying oci/cas/dir implementation
(just for safety). This adds effectively no overhead since
VerifiedReadCloser will not double-up on verification if the digest is
the same. We also add a basic verification of this as an integration and
some unit tests -- so that we can have some confidence it actually works
to protect against bad blobs.

This is an improvement to the hardened VerifiedReadCloser, such that it
now also acts as a verified LimitedReader. First of all, it actually
means we now check the length of blobs according to their descriptors.
It also allows us to avoid reading further than strictly necessary, if
we already know that the read will fail. Unfortunately, the standard
cas/dir cannot really handle this new verification properly.

Fixes #278 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+## Added
+- Full-stack verification of blob hashes and descriptor sizes is now done on
+  all operations, improving our hardening against bad blobs (we already did
+  some verification of layer DiffIDs but this is far more thorough).
+  openSUSE/umoci#278 openSUSE/umoci#280 openSUSE/umoci#282
 
 ## [v0.4.3] - 2018-11-11
 ## Added
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,8 +39,9 @@ RUN zypper -n in \
 
 ENV GOPATH /go
 ENV PATH $GOPATH/bin:$PATH
-RUN go get -u golang.org/x/lint/golint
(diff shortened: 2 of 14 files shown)


cyphar:
LGTM.

tych0:
Just for my own edification, what's the threat that this protects against
that just using DiffIDs doesn't?

tych0:
On Fri, Dec 28, 2018 at 09:24:20AM -0800, Aleksa Sarai [see §317C(6)] wrote:
> A few things:
> 
> 1. It checks the digests of everything (not just layer archives -- JSON blobs are included).

Oh, yeah, that does seem good :)

> 2. In the case of archives, it protects against having a different archive that has the same uncompressed form (DiffIDs are the hash of the uncompressed payload -- meaning you have to uncompresss it in order to check it and also opens you up to archive bombs but that's only partially protected against here).

Yeah, I have to be honest that I've never really understood the point
of the DiffID vs. just hashing the layer itself.

Anyway, thanks for the heads up.

PR #283, opened 2019-01-28
pkg: system: use AT_FDCWD in UtimesNanoAt call

tklauser:
Use unix.AT_FDCWD and the absolute path for unix.UtimesNanoAt in order
to avoid having to open the parent directory on each call to Lutimes.

Signed-off-by: Tobias Klauser <tklauser@distanz.ch>

### DIFF
--- a/pkg/system/utime_linux.go
+++ b/pkg/system/utime_linux.go
@@ -19,10 +19,8 @@ package system
 
 import (
 	"os"
-	"path/filepath"
 	"time"
 
-	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


tklauser:
Ah yes, there even was a comment about it: https://github.com/openSUSE/umoci/blob/0d3b4bba2c22ed0892d7dd762efbf50292a08a7e/pkg/system/utime_linux.go#L35-L39 Thanks!

cyphar:
I wonder if there's now a wrapper for that code. I'll take a look later.

PR #286, opened 2019-02-06
test: replace incorrect use of BATS_TMPDIR

cyphar:
The code that used $BATS_TMPDIR was written before I noticed that it's
actually not a per-test random directory and is instead a way of
referencing /tmp. In particular, the use of $BATS_TMPDIR for $IMAGE was
a blocker for using the new parallel job support I'm working on adding
to bats.

In order to make this not too repetitive, I've created a new
$UMOCI_TMPDIR which is randomised for each test run.


cyphar:
LGTM.

PR #287, opened 2019-02-06
test: use bats --jobs support to speed up tests 

cyphar:
There were a few changes to run() made in bats 1.0, and these broke some
incorrectly-written tests. Fixing this was mostly a few trivial cases of
using "@{foo[@]}" where we weren't previously.

Use the [*not-yet-merged* parallelism support I've been working on in
bats][1] in order to speed up our tests massively. This requires that we
use my home project on OBS, but hopefully my patches will be merged
eventually.

[1]: https://github.com/bats-core/bats-core/pull/172


cyphar:
LGTM.

PR #288, opened 2019-02-06
go.mod: update dependencies to newest tags

cyphar:
* github.com/hashicorp/go-multierror@v1.0.0
 * github.com/klauspost/pgzip@v1.2.1
 * github.com/sirupsen/logrus@v1.3.0
 * github.com/syndtr/gocapability@d98352740cb2
 * github.com/vbatts/go-mtree@v0.4.4
 * golang.org/x/net@d26f9f9a57f3

Signed-off-by: Aleksa Sarai <asarai@suse.de>

PR #289, opened 2019-02-08
Hoist api functions

tych0:
We'd like to use mostly the same code from some of the umoci commands, so let's make them available from some form of "API".

### DIFF
--- a/repack.go
+++ b/repack.go
@@ -1,0 +1,107 @@
+package umoci
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/apex/log"
+	"github.com/openSUSE/umoci/mutate"
+	"github.com/openSUSE/umoci/oci/casext"
+	"github.com/openSUSE/umoci/oci/layer"
+	"github.com/openSUSE/umoci/pkg/fseval"
+	"github.com/openSUSE/umoci/pkg/mtreefilter"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/vbatts/go-mtree"
+	"golang.org/x/net/context"
+)
+
+// Repack repacks a bundle into an image adding a new layer for the changed
+// data in the bundle.
+func Repack(engineExt casext.Engine, tagName string, bundlePath string, meta Meta, history *ispec.History, maskedPaths []string, refreshBundle bool, mutator *mutate.Mutator) error {
+	mtreeName := strings.Replace(meta.From.Descriptor().Digest.String(), ":", "_", 1)
+	mtreePath := filepath.Join(bundlePath, mtreeName+".mtree")
+	fullRootfsPath := filepath.Join(bundlePath, layer.RootfsName)
+
+	log.WithFields(log.Fields{
+		"bundle": bundlePath,
+		"rootfs": layer.RootfsName,
+		"mtree":  mtreePath,
+	}).Debugf("umoci: repacking OCI image")
+
+	mfh, err := os.Open(mtreePath)
+	if err != nil {
+		return errors.Wrap(err, "open mtree")
+	}
+	defer mfh.Close()
+
+	spec, err := mtree.ParseSpec(mfh)
+	if err != nil {
+		return errors.Wrap(err, "parse mtree")
+	}
+
+	log.WithFields(log.Fields{
+		"keywords": MtreeKeywords,
+	}).Debugf("umoci: parsed mtree spec")
+
+	fsEval := fseval.DefaultFsEval
+	if meta.MapOptions.Rootless {
+		fsEval = fseval.RootlessFsEval
+	}
+
+	log.Info("computing filesystem diff ...")
+	diffs, err := mtree.Check(fullRootfsPath, spec, MtreeKeywords, fsEval)
+	if err != nil {
+		return errors.Wrap(err, "check mtree")
+	}
+	log.Info("... done")
+
+	log.WithFields(log.Fields{
+		"ndiff": len(diffs),
+	}).Debugf("umoci: checked mtree spec")
+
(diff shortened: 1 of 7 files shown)


tych0:
I saw various incantations of this error when I was building locally until I moved outside of $GOPATH... not sure what the deal is in travis.

tych0:
Basically what's happening is, go modules is deleting the vendor/ dir, and then the hashes don't match. I'm not sure what the point of these checks is at all, though.

cyphar:
The CI issues should've been fixed by #288, and there were a mix of problems:

 * Travis decided to switch away from GOPATH even though not everyone uses it yet -- and this broke a bunch of things.
 * The issue with `vendor/` was related to `go mod vendor` changing its output based on whether I had a copy of the old `vendor/` tree while doing it.

The reason why we have checks for `vendor/` is because that is used by distributions to build `umoci` -- and so it must be identical to the checked out versions in `go.sum`. An example where I almost missed checking this is [this PR](https://github.com/openSUSE/umoci/pull/279#issuecomment-447692124) -- and so I decided to automate the check.

All of that being said, I'm not sure why it's failing now with your PR. I will look into that. On the PR-side of things, this is something I've wanted for quite a while. :D

cyphar:
#290 should fix the issue.

unpack.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Unpack the runtime bundle.
+	if err := os.MkdirAll(bundlePath, 0755); err != nil {
+		return errors.Wrap(err, "create bundle path")
+	}
+	// XXX: We should probably defer os.RemoveAll(bundlePath).
+
+	// FIXME: Currently we only support OCI layouts, not tar archives. This
+	//        should be fixed once the CAS engine PR is merged into
+	//        image-tools. https://github.com/opencontainers/image-tools/pull/5
```
cyphar: Can you remove this comment?

repack.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/openSUSE/umoci/pkg/mtreefilter"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/vbatts/go-mtree"
+	"golang.org/x/net/context"
+)
+
+// Repack repacks a bundle into an image adding a new layer for the changed
+// data in the bundle.
+func Repack(engineExt casext.Engine, tagName string, bundlePath string, meta Meta, history *ispec.History, maskedPaths []string, refreshBundle bool, mutator *mutate.Mutator) error {
```
cyphar: I really don't like this function signature, but I don't really want to block this PR over this (in my view, the API should be more builder-like than this). If I merge it like this and change it later, would that be more of a pain to you than if we iron out the API now?

tych0:
On Fri, Feb 08, 2019 at 05:16:10PM -0800, Aleksa Sarai [see §317C(6)] wrote:
> cyphar commented on this pull request.
> 
> 
> 
> > +
> +	"github.com/apex/log"
> +	"github.com/openSUSE/umoci/mutate"
> +	"github.com/openSUSE/umoci/oci/casext"
> +	"github.com/openSUSE/umoci/oci/layer"
> +	"github.com/openSUSE/umoci/pkg/fseval"
> +	"github.com/openSUSE/umoci/pkg/mtreefilter"
> +	ispec "github.com/opencontainers/image-spec/specs-go/v1"
> +	"github.com/pkg/errors"
> +	"github.com/vbatts/go-mtree"
> +	"golang.org/x/net/context"
> +)
> +
> +// Repack repacks a bundle into an image adding a new layer for the changed
> +// data in the bundle.
> +func Repack(engineExt casext.Engine, tagName string, bundlePath string, meta Meta, history *ispec.History, maskedPaths []string, refreshBundle bool, mutator *mutate.Mutator) error {
> 
> I really don't like this function signature, but I don't really want to block this PR over this (in my view, the API should be more builder-like than this).

What does more "builder-like" mean? I thought about doing an arguments
struct, but then the benefit of copying and pasting the original code
with the same names won out vs. that. However, that's kind of lame
reason, and I'd be happy to switch it to that, if that's what
builder-like means.

> If I merge it like this and change it later, would that be more of a pain to you than if we iron out the API now?

Either way is fine with me, I'm not in a rush since we're vendoring
this sort of stuff anyway.

cyphar:
> However, that's kind of lame reason, and I'd be happy to switch it to that, if that's what builder-like means.

That would be an improvement to the argument-bake we have here, but builder-like would be somewhat like the proposed `fsconfig(2)` (as an analogy). You would be able to split up some of the pieces of the operation (but provide a more semantic view than the CAS operations themselves) and then "commit" the operation.

But if you are just vendoring, I can always fix this in a future version (and link you to the PR when I've got it so you know how to migrate).

tych0:
On Sat, Feb 09, 2019 at 03:22:05AM -0800, Aleksa Sarai [see §317C(6)] wrote:
> > However, that's kind of lame reason, and I'd be happy to switch it to that, if that's what builder-like means.
> 
> That would be an improvement to the argument-bake we have here, but builder-like would be somewhat like the proposed `fsconfig(2)` (as an analogy). You would be able to split up some of the pieces of the operation (but provide a more semantic view than the CAS operations themselves) and then "commit" the operation.

Ah, I see.

> But if you are just vendoring, I can always fix this in a future version (and link you to the PR when I've got it so you know how to migrate).

Sure, sounds good.

cyphar:
LGTM (testing that lgtm.cyphar.com is working again).

tych0:
Ping, any word on this?

tych0:
Ping, any word on this?

cyphar:
Sorry, this slipped my notice (and I thought we'd decided to fix it before merging). I'm fine with merging it and fixing it later, so I'll merge this. I'll just rebase it for you.

cyphar:
LGTM.

PR #290, opened 2019-02-09
hack: ignore vendor/modules.txt when comparing

cyphar:
It turns out that vendor/modules.txt changes its contents seemingly at
random, and so there's no point including it in our comparison
(especially since it's not used to build anything).


cyphar:
LGTM.

PR #291, opened 2019-03-30
hack: release: make version and arch configurable

cyphar:
In order to be able to handle projects where the version is generated
differently (such as autotools), and using Go for determining the
architecture makes no sense move the generation of those variables to
the user-configurable section.


cyphar:
LGTM.

PR #293, opened 2019-04-02
No empty layers

tych0:
Here's some code to avoid inserting empty layers, and instead just creating a history entry to log that something happened.

### DIFF
--- a/test/repack.bats
+++ b/test/repack.bats
@@ -231,7 +231,7 @@ function teardown() {
 	bundle-verify "$BUNDLE"
 
 	# Make some small change.
-	touch "$BUNDLE/a_small_change"
+	touch "$ROOTFS/a_small_change"
 	now="$(date --iso-8601=seconds --utc)"
 
 	# Repack the image, setting history values.
--- a/test/repack.bats
+++ b/test/repack.bats
@@ -790,3 +790,24 @@ function teardown() {
 	[ "$numLinesB" -gt "$numLinesA" ]
 	[ "$numLinesC" -gt "$numLinesB" ]
 }
+
+@test "umoci repack (empty diff)" {
+	# Unpack the original image
+	new_bundle_rootfs
+	umoci unpack --image "${IMAGE}:${TAG}" "$BUNDLE"
+	[ "$status" -eq 0 ]
+	bundle-verify "$BUNDLE"
+
+	# Repack the image under a new tag.
+	umoci repack --image "${IMAGE}:${TAG}-new" "$BUNDLE"
+	[ "$status" -eq 0 ]
+	image-verify "${IMAGE}"
+
+	# The two manifests should have the same number of layers.
+	manifest0=$(cat "${IMAGE}/oci/index.json" | jq -r .manifests[0].digest | cut -f2 -d:)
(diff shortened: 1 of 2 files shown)


test/repack.bats
```diff
@@ -231,7 +231,7 @@ function teardown() {
 	bundle-verify "$BUNDLE"
 
 	# Make some small change.
-	touch "$BUNDLE/a_small_change"
+	touch "$BUNDLE/rootfs/a_small_change"
```
cyphar: Use `$ROOTFS` -- I thought I fixed all of these incorrect `$BUNDLE` usages...

test/repack.bats
```diff
@@ -790,3 +790,24 @@ function teardown() {
...
+	[ "$status" -eq 0 ]
+	image-verify "${IMAGE}"
+
+    # The two manifests should have the same number of layers.
+    manifest0=$(cat "${IMAGE}/oci/index.json" | jq -r .manifests[0].digest | cut -f2 -d:)
+    manifest1=$(cat "${IMAGE}/oci/index.json" | jq -r .manifests[1].digest | cut -f2 -d:)
+
+    layers0=$(cat "${IMAGE}/oci/blobs/sha256/$manifest0" | jq -r .layers)
+    layers1=$(cat "${IMAGE}/oci/blobs/sha256/$manifest1" | jq -r .layers)
+    [ "$layers0" == "$layers1" ]
```
cyphar: Inconsistent tabs and spaces -- use tabs everywhere.

test/repack.bats
```diff
@@ -790,3 +790,24 @@ function teardown() {
 	[ "$numLinesB" -gt "$numLinesA" ]
 	[ "$numLinesC" -gt "$numLinesB" ]
 }
+
+@test "umoci repack (empty diff)" {
+	# Unpack the original image
+	new_bundle_rootfs && BUNDLE_A="$BUNDLE" ROOTFS_A="$ROOTFS"
```
cyphar: If you're not going to use `BUNDLE_A` and `ROOTFS_A` you can drop the `&& ...` section of this line.

tych0:
Huh. No idea why this one failed. Maybe a race? There's lots of goroutines in that code path...

cyphar:
Yeah, I'm not sure. Every once in a while we see a flaky test (and this one was odd -- a return code was 0 when it shouldn't have been). At some point I should run some stress-like tests to see what falls out.

Anyway, I re-ran it and it passes. I'll review this one tomorrow (the main thing that strikes me is that `mutator.Set` appears to be quite ugly to use if you are changing only one piece of the metadata -- sucks there's no `Option` typeclass in Go and we'd need to do that with pointers).

tych0:
Ping, any word on this?

tych0:
Ping, any word on this?

cyphar:
Yup sorry, this looks okay. I might prefer having a separate `config.go` which meant we didn't need to have a copy of this logic here but I don't mind either way. For OCIv2 (which I have started working on again) we will need to restructure quite a bit of the wrappers so that the same interfaces can support both layered rootfs-es (OCIv1) and tree-based rootfs-es (OCIv2).

LGTM.

cyphar:
I think I've figured out the root cause of the failure. I think it's because the current trick of returning errors through `write.CloseWithError` seems to have some sort of race condition involved...

cyphar:
LGTM.

Issue #294, opened 2019-04-06
Rootless config generates invalid mount for /etc/resolv.conf


rneilson:
Using `umoci unpack --rootless` (or `umoci raw config --rootless`), an extra mount is generated for `/etc/resolv.conf`, with `"type": "none"` [1](https://github.com/openSUSE/umoci/blob/master/oci/layer/unpack.go#L356). This causes `runc` to parse the mount with the default (directory) behavior [2](https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/spec_linux.go#L279), trying to create the destination as a directory instead of a file [3](https://github.com/opencontainers/runc/blob/v1.0.0-rc6/libcontainer/rootfs_linux.go#L387).

TBH, not sure if this is a runtime spec problem, a `runc` mount parsing problem, or a `umoci` config generation problem, but I'm guessing it's most easily fixed here.

To reproduce:
```
> skopeo copy docker://busybox:latest oci:busybox:latest
> umoci raw unpack --rootless --image busybox rootfs
> umoci raw config --image=busybox --rootless --rootfs=rootfs config.json
> runc run busybox
container_linux.go:344: starting container process caused "process_linux.go:424: container init caused \"rootfs_linux.go:58: mounting \\\"/etc/resolv.conf\\\" to rootfs \\\"/data/contained/temp/rootfs\\\" at \\\"/data/contained/temp/rootfs/etc/resolv.conf\\\" caused \\\"not a directory\\\"\""
```

`umoci --version`
```
umoci version 0.4.4
```

`runc --version `
```
runc version 1.0.0-rc6
spec: 1.0.1-dev
```

cyphar:
The `resolv.conf` mount is usually necessary, but I think the issue might be that the rootfs doesn't already have a `resolv.conf` and so runc gets a bit confused? I will take a look at this...

rneilson:
I tried adding an empty `resolv.conf` to the rootfs as an empty file before running the container, and I get another, related error instead:
```
container_linux.go:344: starting container process caused "process_linux.go:424: container init caused \"rootfs_linux.go:58: mounting \\\"/etc/resolv.conf\\\" to rootfs \\\"/data/contained/unpacked/busybox/rootfs\\\" at \\\"/data/contained/unpacked/busybox/rootfs/etc/resolv.conf\\\" caused \\\"mkdir /data/contained/unpacked/busybox/rootfs/etc/resolv.conf: not a directory\\\"\""
```

[...]

PR #295, opened 2019-04-08
layer: specify "type: bind" in config.json for bind-mounts

cyphar:
Due to an oversight in runc, setting a type other than "" or "bind" will
result in weird errors (especially when dealing with file-based
bind-mounts).

To get around this, we can simply set an empty type. This oversight is
being dealt with upstream, but won't help old runc releases (so we need
this workaround -- and we use "bind" to work with much older releases as
well).

Fixes: #294


cyphar:
LGTM. I've tested it with the reproducer and it works now.

cyphar:
Long-term we would want to use `runc` in the integration tests. I will look into this.

PR #296, opened 2019-04-08
test: add runc smoke-test

cyphar:
We've had cases where the default setup of "umoci + runc" didn't work
(usually in rootless container scenarios). Since this is a very common
usage of umoci, it makes sense to check it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ RUN zypper -n in \
 		oci-image-tools \
 		oci-runtime-tools \
 		python-setuptools python-xattr attr \
+		runc \
 		skopeo \
 		tar
 RUN useradd -u 1000 -m -d /home/rootless -s /bin/bash rootless
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,25 @@ CMD := ${PROJECT}/cmd/umoci
 
 # We use Docker because Go is just horrific to deal with.
 UMOCI_IMAGE := umoci_dev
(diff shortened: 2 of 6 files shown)


cyphar:
Failure is making very little sense, but I think it has to do with the context the test is being run in. Strangely, if you call `bats-exec-test` directly there isn't an error inside `runc`. But `bats-exec-suite` triggers it.

```
% bats-exec-test test/runc.bats test_umoci_-2b_runc_-5bsmoke_test-5d
```

cyphar:
Okay, so the issue is (as usual) my fault. As part of the [`-j` support in bats](https://github.com/bats-core/bats-core/pull/172), `bats-exec-test` now runs in a context without stdin. For some reason this causes `runc` to fail (which might also be my fault):

```
% sudo ./runc run -b bundle ctr </dev/null
WARN[0000] exit status 2                    
ERRO[0000] container_linux.go:353: starting container process caused: process_linux.go:450: container init caused: 
```

The temporary workaround is to set `terminal: false`, but I will open a bug report in runc about this.

cyphar:
🧐

```
mount("none", "/", NULL, MS_REC|MS_PRIVATE, NULL) = -1 EACCES (Permission denied)
```

cyphar:
LGTM.

/cc @tych0 @vbatts

tych0:
LGTM.

tych0:
I just pressed github's merge button; let me know if I should be doing that differently :)

cyphar:
That's perfectly fine. I personally prefer to use `git merge --no-ff` so that I can sign my merge commits, but it's totally up to you.

PR #297, opened 2019-05-16
After layer callback

tych0:
We have code that wants to do a "partial" extraction of certain layers in the image, so let's add some hooks to do that.


cyphar:
I do like this idea (and is something I've been planning to work on for quite a while). Unfortunately I'm not a huge fan of this interface (and it's also adding even more arguments to the super-unwieldy `UnpackLayer`). I will give this a more in-depth review next week and hopefully come up with a suggestion that works for both of us. (My long-term goal would be to be able to have hooks running on the command-line so you could construct your own storage drivers around umoci with shell scripts.)

tych0:
Yes, I agree that exposing the components in a per-layer way would be much better. I didn't think very hard about this API design, I just did something that was small and worked. I figured you were going to change all the UnpackRootfs() type stuff soon anyway, so I'd just do something easy :). Anyway, if you have a better implementation, I'm happy to rebase our stuff on top of it.

tych0:
Ping - any thoughts on this?

cyphar:
So, I'm *okay* with this PR as-is, but I would like to change how this interface works in the future. My long-term goal for hooks within `umoci` is that we should be able to have most of the extraction process be hookable (and configurable with a config file) so that. The Go API would also need to change a fair bit for this (probably taking `map[HookPoint]callback` or something).

I also think that `startsFrom` shouldn't be a separate argument -- it should be a special return value (like `filepath.Walk`) from a pre-extraction hook which indicates "skip this layer". Then, you can implement `startsFrom` in the hook (as well as for instance implementing skipping of any non-distributable layers or whatever else you like).

But, I'd be okay with merging this (after a rebase) and fixing up the API later as long as you're alright with the API changing underneath you in a future release.

cyphar:
Oh, and sorry for letting this one languish for so long. I really have been putting too many things on my plate for the past two years or so.

cyphar:
LGTM.

PR #304, opened 2019-10-16
tar_extract: don't error on fs without xattr support

### LINKED ISSUE #303: ENOTSUP error in extraction to NFS
In https://github.com/sylabs/singularity/issues/4593 a Singularity user reports trying to build a sandbox (directory) rootfs to an NFS mounted filesystem that doesn't support xattrs. We use umoci to extract the layers to a rootfs.

This errors out with the following from the umoci code:

```
get dirHdr.Xattrs: unpriv.llistxattr: operation not supported
```

This seems to be triggered in the code here:

https://github.com/openSUSE/umoci/blob/4fc9627cfc26f9e2a2af67662cb1e3381d4580f6/oci/layer/tar_extract.go#L307

There is an `Llistxattr` which is not protected against `ENOTSUP` on fs that do not handle xattrs in the same way as other xattr operations in the `tar_extract.go` code.

Is it reasonable to just add a warning similar to:

https://github.com/openSUSE/umoci/blob/4fc9627cfc26f9e2a2af67662cb1e3381d4580f6/oci/layer/tar_extract.go#L178

... here?

dctrud:
This is a simple proposal to address #303 - I understand this may be naive / it may be that inability to deal with the `xattrs` is considered more than a warning.

Closes #303

If we are extracting to a filesystem that does not support xattrs, make
sure that an ENOTSUP from clearxattr or listxattr results in a warning,
not an error.

### DIFF
--- a/oci/layer/tar_extract.go
+++ b/oci/layer/tar_extract.go
@@ -158,11 +158,8 @@ func (te *TarExtractor) restoreMetadata(path string, hdr *tar.Header) error {
 					continue
 				}
 			}
-			if te.partialRootless {
-				log.Warnf("rootless{%s} ignoring forbidden xattr: %q", hdr.Name, name)
-				continue
-			}
-			return errors.Errorf("restore xattr metadata: saw forbidden xattr %q: %s", name, hdr.Name)
+			log.Warnf("xattr{%s} ignoring forbidden xattr: %q", hdr.Name, name)
+			continue
 		}
 		if err := te.fsEval.Lsetxattr(path, name, value, 0); err != nil {
 			// In rootless mode, some xattrs will fail (security.capability).
--- a/oci/layer/tar_extract.go
+++ b/oci/layer/tar_extract.go
@@ -131,9 +131,14 @@ func (te *TarExtractor) restoreMetadata(path string, hdr *tar.Header) error {
 	// Apply xattrs. In order to make sure that we *only* have the xattr set we
 	// want, we first clear the set of xattrs from the file then apply the ones
 	// set in the tar.Header.
-	if err := te.fsEval.Lclearxattrs(path, ignoreXattrs); err != nil {
-		return errors.Wrapf(err, "clear xattr metadata: %s", path)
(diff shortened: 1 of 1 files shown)


oci/layer/tar_extract.go
```diff
@@ -131,9 +131,14 @@ func (te *TarExtractor) restoreMetadata(path string, hdr *tar.Header) error {
...
 	// want, we first clear the set of xattrs from the file then apply the ones
 	// set in the tar.Header.
-	if err := te.fsEval.Lclearxattrs(path, ignoreXattrs); err != nil {
-		return errors.Wrapf(err, "clear xattr metadata: %s", path)
+	err := te.fsEval.Lclearxattrs(path, ignoreXattrs)
+	if err != nil {
+		if errors.Cause(err) != unix.ENOTSUP {
+			return errors.Wrapf(err, "clear xattr metadata: %s", path)
+		}
+		log.Warnf("ignoring ENOTSUP on clearxattrs %q", path)
```
cyphar: Please follow the same formatting for warnings:

```
log.Warnf("xattr{%s} ignoring ENOTSUP on setxattr %q", hdr.Name, name)
```

(I also just noticed a typo in the original `ENOTSUP` check -- while you're at it, can you please change it from `xatt` to `xattr`? Thanks.)

oci/layer/tar_extract.go
```diff
@@ -306,7 +311,10 @@ func (te *TarExtractor) UnpackEntry(root string, hdr *tar.Header, r io.Reader) (
 		//       tar_generate.go.
 		xattrs, err := te.fsEval.Llistxattr(dir)
 		if err != nil {
-			return errors.Wrap(err, "get dirHdr.Xattrs")
+			if errors.Cause(err) != unix.ENOTSUP {
+				return errors.Wrap(err, "get dirHdr.Xattrs")
+			}
+			log.Warnf("ignoring ENOTSUP on listxattr %q", dir)
```
cyphar: Same nit as above.

PR #305, opened 2019-10-17
tar_extract: only warn for forbidden xattrs

### LINKED ISSUE #302: selinux xattr errors for Scientific Linux as with root
We're now using `umoci` for OCI layer extractions in https://github.com/sylabs/singularity due to the more faithful extraction of permissions compared to `containers/image-tools` - as recommended by @cyphar in https://github.com/opencontainers/image-tools/issues/218

In a recent issue https://github.com/sylabs/singularity/issues/4578 a user reports that umoci extraction errors out for the CERN Scientific Linux 6 docker image (cern/slc6-base:latest). The reason for this is that the layers for this image contain selinux xattrs. In the rootless flow (`--fakeroot` option for Singularity builds) umoci will just warn about the forbidden xattr. In the flow for extraction with root it will error.

```
unpack layer: unpack entry: .: apply hdr metadata: restore xattr metadata: saw forbidden xattr "security.selinux": .
```

I understand from @olifre who reported the issue to Singularity that the CERN SL6 docker image is built with koji. Following up the trail I see that koji is using https://github.com/redhat-imaging/imagefactory/commit/8c61f3e2c494771a23f2649360c3f6528fbcfe73 to build a docker image. Since the referenced commit in mid 2016 that will be excluding selinux xattrs - so there's an argument here that the docker image is broken. However, given that docker can handle this without issue would it be reasonable for umoci to warn rather than error in extractions as root (like it already does rootless), by default or via an option?

dctrud:
Closes #302

Rootless mode currently warns if a forbidden xattr is seen, while
extractions as root error out. Make root extractions warn, so that
docker images such as cern/sl6-base:latest can be extracted as root
without failing due to this error.

### DIFF
--- a/oci/layer/tar_extract.go
+++ b/oci/layer/tar_extract.go
@@ -153,11 +153,8 @@ func (te *TarExtractor) restoreMetadata(path string, hdr *tar.Header) error {
 					continue
 				}
 			}
-			if te.partialRootless {
-				log.Warnf("rootless{%s} ignoring forbidden xattr: %q", hdr.Name, name)
-				continue
-			}
-			return errors.Errorf("restore xattr metadata: saw forbidden xattr %q: %s", name, hdr.Name)
+			log.Warnf("xattr{%s} ignoring forbidden xattr: %q", hdr.Name, name)
+			continue
 		}
 		if err := te.fsEval.Lsetxattr(path, name, value, 0); err != nil {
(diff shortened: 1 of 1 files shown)


cyphar:
Would you mind merging this with #304 (feel free to keep them as separate commits)?

dctrud:
Commit added to #304

PR #306, opened 2019-10-25
tar_extract: don't mask original error on short write

cyphar:
If there was an error in io.Copy() which resulted in a short write, we
used to hide the error (though the error would be quite useful for
debugging). Thus only set err to io.ErrShortWrite if there's no error
set already.

Fixes #301
Reported-by: Tycho Andersen <tycho@tycho.ws>


tych0:
LGTM, thanks!

cyphar:
LGTM.

PR #307, opened 2019-10-30
oci: casext: further hookification of blobs

cyphar:
This is necessary in order for references to non-stock OCI blobs to work
properly (in order for OCIv2 to work properly). Hopefully this is also
extensible enough that anyone who has a custom blob format can make use
of it.


cyphar:
LGTM.

PR #311, opened 2020-01-27
tar_extract: only warn once about xattr ENOTSUP

dctrud:
When a destination filesystem does not support xattrs we are rasing at
least 1 ENOTSUP related warning per file. This is unhelpful, and can
scroll the terminal obscuring more important warnings. Instead, show a
single warning for the first occurrence, with an improved explanation of
the cause.

### DIFF
--- a/oci/layer/tar_extract.go
+++ b/oci/layer/tar_extract.go
@@ -64,6 +64,13 @@ type TarExtractor struct {
 	// to a trie if necessary). These paths are relative to the tar root but
 	// are fully symlink-expanded so no need to worry about that line noise.
 	upperPaths map[string]struct{}
+
+	// enotsupWarned is a flag set when we encounter the first ENOTSUP error
+	// dealing with xattrs. This is used to ensure extraction to a destination
+	// file system that does not support xattrs raises a single warning, rather
+	// than a warning for every file, which can amount to 1000s of messages that
+	// scroll a terminal, and may obscure other more important warnings.
+	enotsupWarned bool
 }
 
 // NewTarExtractor creates a new TarExtractor.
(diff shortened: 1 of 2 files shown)


oci/layer/tar_extract.go
```diff
@@ -136,7 +144,11 @@ func (te *TarExtractor) restoreMetadata(path string, hdr *tar.Header) error {
 		if errors.Cause(err) != unix.ENOTSUP {
 			return errors.Wrapf(err, "clear xattr metadata: %s", path)
 		}
-		log.Warnf("xattr{%s} ignoring ENOTSUP on clearxattrs", path)
+		if !te.enotsupWarned {
+			log.Warnf("xattr{%s} ignoring ENOTSUP on clearxattrs", path)
+			log.Warnf("xattr{%s} destination filesystem does not support xattrs, further warnings will be suppressed", path)
+			te.enotsupWarned = true
```
cyphar: This patch does make sense, my only comment is that this behaviour should be disabled if the logging level is debug.

oci/layer/tar_extract.go
```diff
@@ -136,7 +144,13 @@ func (te *TarExtractor) restoreMetadata(path string, hdr *tar.Header) error {
...
+		} else {
```
dctrud: Is it sufficient to drop a `Debugf` into an `else` for the repeated warnings as done here - or would you prefer everything to come out as a warning (original behavior) if the loglevel is debug?

oci/layer/tar_extract.go
```diff
@@ -136,7 +144,13 @@ func (te *TarExtractor) restoreMetadata(path string, hdr *tar.Header) error {
...
+			log.Debugf("xattr{%s} ignoring ENOTSUP on clearxattrs", path)
```
dctrud: Is it sufficient to drop a `Debugf` into an `else` for the repeated warnings as done here - or would you prefer everything to come out as a warning (original behavior) if the loglevel is debug?

cyphar: No, I think it's alright to have them be debug messages.

cyphar:
LGTM.

PR #314, opened 2020-03-16
gc: drive gc using digests instead of image tags/names

rchincha:
garbage collection enforces an implicit policy that only images with
names and tags must be kept and the rest is fair game to prune.
However, as per the OCI spec [1], "annotations" field is optional.

This path uses only digests to determine reachability.
If a particular image needs to be garbage collected then remove the
image from the index.json and run gc.

Note that this is a behavior change in that some images which were
pruned earlier will now not be.

Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>

References:
[1] https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions


oci/casext/refname.go
```diff
@@ -206,10 +198,8 @@ func (e Engine) ListReferences(ctx context.Context) ([]string, error) {
 
 	var refs []string
 	for _, descriptor := range index.Manifests {
-		ref, ok := descriptor.Annotations[ispec.AnnotationRefName]
-		if ok {
-			refs = append(refs, ref)
-		}
+		ref := descriptor.Digest.String()
```
tych0: This will break a lot of existing code and doesn't seem generally necessary to accomplish the goal in the commit description?

cyphar: It's also semantically incorrect -- the point of this method is to give you the list of all reference names in the image. A list of digests (without blob type or any other important information) is not equivalent.

oci/casext/refname.go
```diff
@@ -76,8 +69,7 @@ func (e Engine) ResolveReference(ctx context.Context, refname string) ([]Descrip
 	// the time of writing -- 1.0.0-rc5) there are some discussions to add this
 	// restriction in 1.0.0-rc6.
 	for _, descriptor := range index.Manifests {
-		// XXX: What should we do if refname == "".
-		if descriptor.Annotations[ispec.AnnotationRefName] == refname {
+		if descriptor.Digest.String() == refname {
```
cyphar: This change is also incorrect -- it breaks basically all of the `umoci` command-line tools because now reference resolution requires you to specify the digest (this code is what actually figures out what `ubuntu:latest` means in an OCI image).

cyphar:
> garbage collection enforces an implicit policy that only images with
names and tags must be kept and the rest is fair game to prune.
However, as per the OCI spec [1], "annotations" field is optional.

It is, but in `index.json` there is a special meaning for the `org.opencontainers.image.ref.name`. That's how the set of "image" tags are defined in OCI. So it's not wrong to do it this way (and the OCI spec doesn't define a "garbage collection" interface, so we're really free to define our own semantics).

However, I can see an argument for the current semantics being less-than-ideal. If you have a descriptor in the `index.json`, really it (and all its children) should be included in the "root set" and not garbage collected when we sweep.

But this patch is the wrong way of implementing this (and it breaks a bunch of other code, to boot). Instead, you should change the garbage collection algorithm in `oci/casext/gc.go`. Rather than using `ListReferences`, we should be doing `GetIndex` and then iterating through the list of manifests (the rest of the `e.Reachable` logic should follow cleanly from there).

rchincha:
> 
> 
> See my comment.

@cyphar , @tych0 
Thanks for the review, will update the PR.

cyphar:
LGTM, though it might be a good idea to add a test for this.

tych0:
LGTM as well, and yes, a test would be appreciated :)

rchincha:
@tych0, @cyphar 

Will add a test case for this.

Just to be clear, this change is not based off some hypothetical.
As per the dist-spec [1], it is indeed possible to upload blobs and then create a manifest via its digest alone (the other option being name/tag).

[1] https://github.com/opencontainers/distribution-spec/blob/master/spec.md#put-manifest

cyphar:
Sure, though distribution-spec has a bit of an "odd" relationship with the image-spec -- in particular `index.json` isn't really the same thing as the manifest lists you get with distribution-spec. In fact, the `/v2/<name>/manifests/<reference>` endpoint arguably should be represented as a list of manifests in `index.json` that have `org.opencontainers.image.refs.name` equal to `<reference>` (or whatever) -- because "references" don't exist in OCI images but do exist in distribution-spec.

Aren't containers fun...

rchincha:
=== RUN   TestGC

2020/03/19 00:02:28  info garbage collecting blob: sha256:005d05821bcc432915b56b1681f8ed4035818233edc3d2b79f2916cb2c9978e3

2020/03/19 00:02:28  info garbage collecting blob: sha256:02e5ac821931a2a461c30ce10a47c7dd6857ceb2c0371d4d5f431cdd4b02d25e

...

2020/03/19 00:02:28  info garbage collecting blob: sha256:fb415ae9b5809d6b34b0220e2517dcf8cf084412b4b1c253a41e5242391d1970

2020/03/19 00:02:28  info garbage collecting blob: sha256:fba732be9aa7ed555c192f482f1c2459834be27432c8f00cff9b2a041f6de522

--- PASS: TestGC (1.21s)

oci/casext/gc_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	if descMap == nil {
+		t.Fatalf("empty descMap")
+	}
+
+	err = engineExt.GC(ctx)
+	if err != nil {
+		t.Fatalf("GC failed: %+v", err)
+	}
```
tych0: Thanks for the test. Should we add something that verifies that GC() actually did the right thing, not just that it didn't return an error?

rchincha: Will look into it. Still learning the code as I go.

rchincha:
Updated the PR.

tych0:
Great, thanks. @cyphar should I just press github merge? Looks like you have a fancier way of doing it?

cyphar:
Since I sign all of my commits, I manually merge PRs so I can sign the merge commit. But feel free to do it however you like. But since I'm here, I'll do the merge. :wink:

cyphar:
For reference, I don't do anything too fancy. With my `git` aliases, I just do:

```
% git pr 314 # fetch the PR 314 into a branch called "pr-314"
% git mm 314 # git merge --no-ff to get a merge commit
% git push origin master
```

And `git pr $n` is just an alias for `git fetch origin pull/$n/head:pr-$n` which on GitHub gives you the PR branch.

tych0:
And it looks like you manually write the commit message?

cyphar:
Yeah, I do clean it up a bit but that's just a habit I got into (the one thing that bothered me with GitHub merges with LGTMs-in-comments was that you don't have a history of reviewers like you do in Linux). But I don't really mind how you choose to merge things -- it's just my OCD playing up. :wink:

tych0:
Ok, sounds good, thanks!

PR #315, opened 2020-03-25
api: allow for custom filters in Repack()

tych0:
Instead of just allowing for masking paths, let's allow for completely
custom filters in Repack().

### DIFF
--- a/cmd/umoci/repack.go
+++ b/cmd/umoci/repack.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openSUSE/umoci/oci/cas/dir"
 	"github.com/openSUSE/umoci/oci/casext"
 	igen "github.com/openSUSE/umoci/oci/config/generate"
+	"github.com/openSUSE/umoci/pkg/mtreefilter"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
@@ -171,5 +172,9 @@ func repack(ctx *cli.Context) error {
 		}
 	}
 
-	return umoci.Repack(engineExt, tagName, bundlePath, meta, history, maskedPaths, ctx.Bool("refresh-bundle"), mutator)
+	filters := []mtreefilter.FilterFunc{
+		mtreefilter.MaskFilter(maskedPaths),
+	}
+
+	return umoci.Repack(engineExt, tagName, bundlePath, meta, history, filters, ctx.Bool("refresh-bundle"), mutator)
 }
--- a/repack.go
+++ b/repack.go
@@ -36,7 +36,7 @@ import (
 
 // Repack repacks a bundle into an image adding a new layer for the changed
 // data in the bundle.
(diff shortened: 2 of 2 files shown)


cyphar:
LGTM, this seems reasonable. Though, to be honest I've never liked the whole `mtreefilter` thing.

repack.go
```diff
@@ -78,9 +78,8 @@ func Repack(engineExt casext.Engine, tagName string, bundlePath string, meta Met
 		"ndiff": len(diffs),
 	}).Debugf("umoci: checked mtree spec")
 
-	diffs = mtreefilter.FilterDeltas(diffs,
-		mtreefilter.MaskFilter(maskedPaths),
-		mtreefilter.SimplifyFilter(diffs))
+	allFilters := append([]mtreefilter.FilterFunc{mtreefilter.SimplifyFilter(diffs)}, filters...)
```
cyphar: Did you mean to put `SimplifyFilter` at the end as it was originally? In which case it should be

```
filters = append(filters, mtree.SimplifyFilter(diffs))`
```

tych0: No, I meant to put it at the beginning; if it's the first filter run, I guess there will be less overall filtering needed? I can move it to the end if you want, though.

cyphar: The filters will be run the same number of times (each filter is run for each element -- maybe we shouldn't do it that way but :man_shrugging:), but I don't really mind either way. I just think

```
filters = append(filters, mtree.SimplifyFilter(diffs))
```

is slightly less ugly.

cyphar:
LGTM aside from my nit (which you can just ignore).

repack.go
tych0: Oh, ok. I'll switch it then, I figured it chained them all. Agreed it's less ugly :)

cmd/umoci/repack.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/openSUSE/umoci/mutate"
 	"github.com/openSUSE/umoci/oci/cas/dir"
 	"github.com/openSUSE/umoci/oci/casext"
+	"github.com/openSUSE/umoci/pkg/mtreefilter"
 	igen "github.com/openSUSE/umoci/oci/config/generate"
```
cyphar: `go fmt` doesn't like this.

tych0:
On Thu, Mar 26, 2020 at 09:37:57AM -0700, Aleksa Sarai [see §317C(6)] wrote:
> @cyphar commented on this pull request.
> 
> 
> 
> > @@ -26,6 +26,7 @@ import (
>  	"github.com/openSUSE/umoci/mutate"
>  	"github.com/openSUSE/umoci/oci/cas/dir"
>  	"github.com/openSUSE/umoci/oci/casext"
> +	"github.com/openSUSE/umoci/pkg/mtreefilter"
>  	igen "github.com/openSUSE/umoci/oci/config/generate"
> 
> `go fmt` doesn't like this.

Arg, I guess that's not part of the test suite? Anyway, fixed, thanks.

cyphar:
LGTM. The `go fmt` check is done as part of `make ci` (more specifically, `make validate`).

PR #317, opened 2020-05-11
unpack: clean up error message

tych0:
Right now this is something like:

error: config.json already exists
bundle path empty

which is silly and confusing. Let's just return the "already exists" as an
error. Additionally, the "bundle path empty" message is wrong, this error
can only occur when there's a problem accessing the bundle path (that is:
it is not an IsNotExist() error), so it's explicitly not empty :)


cyphar:
I didn't see this earlier -- LGTM.

PR #318, opened 2020-05-26
*: get MacOS build working

### LINKED ISSUE #98: OS X Support
Thanks for providing this project.

Are there plans to support OS X?

cyphar:
I don't have a MacOS machine (nor can we currently run the tests on
MacOS) but we can at least do a sanity-check by trying to build the
project on MacOS.

Fixes #98


cyphar:
The only unit test failing now is because apparently MacOS doesn't support having a hardlink to a symlink (not exactly surprising -- though you get `ENOENT` which is a bit of a fruity response to get). So we'll need to adjust tests for that, but the other test adjustments (making the remounting tests Linux-only, and dealing with the second-based time granularity of MacOS filesystems). Integration tests might take some more work to do (though I think most should be fine, the only major issue is there are a probably a fair few Linux-isms that we'll need to work around).

cyphar:
Unit tests are working fine under MacOS now. I am a little worried how we'll deal with `libpathrs` in the future, but we'll cross that bridge when we come to it. Integration tests might take some more work to do (though I think most should be fine, the only major issue is there are a probably a fair few Linux-isms that we'll need to work around).

cyphar:
You can let me know if this works @IronCore864, @huntc.

cyphar:
LGTM.

/cc @tych0 @vbatts

tych0:
LGTM.

PR #328, opened 2020-06-20
umoci config: disable flag argument re-ordering

cyphar:
It turns out that urfave/cli [hasn't been happy with parsing
command-lines like [--config.cmd -c] for quite some time, and only
just exposed the issue in a test-case][1]. There are loads of more subtle
cases where you could see this happened even in the version of
urfave/cli we have vendored.

Until it's fixed upstream (which might not happen), we have to work
around this by disabling argument reordering for 'umoci config' (which
is the only command that usually has to deal with such flag arguments,
and luckily doesn't take any positional arguments).

[1]: https://github.com/urfave/cli/issues/1152


cyphar:
LGTM.

/cc @vbatts @tych0

tych0:
LGTM

PR #338, opened 2020-06-29
GC: add support for GC policies

rchincha:
For our dist-spec implementation [1][2],
we have a use case where we use umoci's GC code to clean up orphaned
blobs. The issue is that as per dist-spec layer upload and manifest
updates are two different API calls and typically in that order, which
means layers begin life as orphans.

Adding the proposed patch allows us to:
a) mitigate the above issue
b) have a generic policy framework for future expansion

[1] https://github.com/opencontainers/distribution-spec
[2] https://github.com/anuvu/zot

### DIFF
--- a/oci/casext/gc.go
+++ b/oci/casext/gc.go
@@ -25,6 +25,9 @@ import (
 	"golang.org/x/net/context"
 )
 
+// GCPolicy is a policy function that returns 'true' if a blob can be GC'ed
+type GCPolicy func(ctx context.Context, digest digest.Digest) (bool, error)
+
 // GC will perform a mark-and-sweep garbage collection of the OCI image
 // referenced by the given CAS engine. The root set is taken to be the set of
 // references stored in the image, and all blobs not reachable by following a
@@ -35,7 +38,11 @@ import (
 // functions. In other words, it assumes it is the only user of the image that
 // is making modifications. Things will not go well if this assumption is
 // challenged.
-func (e Engine) GC(ctx context.Context) error {
+//
+// Furthermore, GC policies (zero or more) can also be specified which given a
+// blob's digest can indicate whether that blob needs to garbage collected. The
+// blob is skipped for garbage collection if a policy returns false.
+func (e Engine) GC(ctx context.Context, policies ...GCPolicy) error {
 	// Generate the root set of descriptors.
 	var root []ispec.Descriptor
 
@@ -74,12 +81,26 @@ func (e Engine) GC(ctx context.Context) error {
 	}
 
 	n := 0
+sweep:
 	for _, digest := range blobs {
 		if _, ok := black[digest]; ok {
 			// Digest is in the black set.
 			continue
 		}
-		log.Infof("garbage collecting blob: %s", digest)
+
+		for i, policy := range policies {
+			ok, err := policy(ctx, digest)
+			if err != nil {
+				return errors.Wrapf(err, "invoking policy %d failed", i)
+			}
+
+			if !ok {
+				// skip this blob for GC
+				log.Debugf("skipping garbage collection of blob %s because of policy %d", digest, i)
+				continue sweep
+			}
+		}
+		log.Debugf("garbage collecting blob: %s", digest)
 
 		if err := e.DeleteBlob(ctx, digest); err != nil {
 			return errors.Wrapf(err, "remove unmarked blob %s", digest)
--- a/oci/casext/gc_test.go
+++ b/oci/casext/gc_test.go
@@ -3,12 +3,14 @@ package casext
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/opencontainers/go-digest"
 	imeta "github.com/opencontainers/image-spec/specs-go"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/umoci/oci/cas/dir"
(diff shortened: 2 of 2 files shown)


oci/casext/gc.go
```diff
@@ -74,11 +77,25 @@ func (e Engine) GC(ctx context.Context) error {
...
 	for _, digest := range blobs {
 		if _, ok := black[digest]; ok {
 			// Digest is in the black set.
 			continue
 		}
+
+		// Iterate over all policy funcs and if any one of them indicates that GC
+		// must not be performed, skip that blob
+		for _, policy := range policies {
+			ok, err := policy(digest)
```
tych0, line 90: I wonder if this couldn't/shouldn't be a descriptor instead of a digest? I've written a lot of code switch()ing on mime types recently, so that would be handy...

rchincha: Had to change the signature of walk.go:Reachable(). Either that, or we need a different method in GC path.

tych0: That's fine, this LGTM, thanks!

tych0:
LGTM

oci/casext/gc.go
cyphar: You don't need to change the signature of `Reachable` (in fact, please don't do that) you can just use `Paths` (which gives you the set of `DescriptorPath`s you can reach -- and `DescriptorPaths` would be even better than `Descriptor`s). Maybe I should've called it `ReachablePaths`...

Given that this is all about adding policy functions, it does make sense to let those functions know what type the object is pointing to (though it is technically possible -- though unlikely -- that the same blob is referenced using different mime types).

oci/casext/walk.go
```diff
@@ -176,19 +176,19 @@ func (e Engine) Paths(ctx context.Context, root ispec.Descriptor) ([]DescriptorP
 // Walk(). The returned slice will *not* contain any duplicate digest.Digest
 // entries. Note that without descriptors, a digest is not particularly
 // meaninful (OCI blobs are not self-descriptive).
-func (e Engine) Reachable(ctx context.Context, root ispec.Descriptor) ([]digest.Digest, error) {
-	seen := map[digest.Digest]struct{}{}
+func (e Engine) Reachable(ctx context.Context, root ispec.Descriptor) ([]ispec.Descriptor, error) {
```
cyphar, line 179: Please don't do this, use `Paths`.

oci/casext/walk.go
```diff
@@ -176,19 +176,19 @@ func (e Engine) Paths(ctx context.Context, root ispec.Descriptor) ([]DescriptorP
...
+	seen := map[digest.Digest]ispec.Descriptor{}
```
cyphar, line 180: And this is incorrect because you can have multiple descriptors (with different values) pointing to the same digest.

oci/casext/gc.go
```diff
@@ -25,6 +25,9 @@ import (
 	"golang.org/x/net/context"
 )
 
+// GCPolicy is a policy function that returns if a blob can be GC'ed
+type GCPolicy func(desc ispec.Descriptor) (bool, error)
```
cyphar, line 29: In line with switching to `Paths`, this should be a `DescriptorPath`.

oci/casext/gc.go
```diff
@@ -35,7 +38,7 @@ import (
 // functions. In other words, it assumes it is the only user of the image that
 // is making modifications. Things will not go well if this assumption is
 // challenged.
```
cyphar, line 40: This doc comment needs to be updated.

oci/casext/gc.go
```diff
@@ -35,7 +38,7 @@ import (
...
-func (e Engine) GC(ctx context.Context) error {
+func (e Engine) GC(ctx context.Context, policies ...GCPolicy) error {
```
cyphar, line 41: I'm wondering whether it'd be better to have this just take a single function -- callers can always write their own logic for multiple policies but we don't know whether a caller would prefer that we require *all* policies to match or just one policy to match. In addition, what if the caller wants all of the policy functions to be executed, even if one said to keep the blob?

But if we do that then we should create a new `GC` function called `GCWithPolicy` (or something) which takes a single function as an argument, and leave the `GC` signature the same.

tych0: Yeah, that all sounds reasonable; GCWithPolicy() would save a lot of code updates everywhere for everyone :)

cyphar: Though in fairness (and I missed this when I first looked) -- the `GC(ctx, ...)` approach actually doesn't break anyone's builds because not passing any extra arguments gives you a `nil` slice.

oci/casext/gc.go
rchincha: @cyphar Made the recommended changes. Would keep the policy funcs as agnostic about umoci's internals as possible, so would stay in the OCI domain.

oci/casext/gc.go
rchincha: Tried to make sure there is no behaviour or API changes as much as possible.

oci/casext/gc.go
rchincha: Want to hide umoci internals from the policy func writers. So just pass in the standard ispec.Descriptor.

oci/casext/gc.go
rchincha: That the blob list is no longer "stable" while making the GC() call? Still true in our use case as we hold a lock when calling GC().

oci/casext/walk.go
rchincha: Hmm...the issue is then is that the policy may "ok" to GC for one descriptor and later "not ok" but it is now too late.

oci/casext/gc.go
cyphar: I mean that you need to update the doc comment to explain what the `policies` arguments do.

oci/casext/gc.go
cyphar: `DescriptorPath` isn't really an umoci internal, it's fairly publicly used if you want to do `ResolveReference` (so basically every umoci API user will see `DescriptorPath` at some point). For a bit of background -- the reason why `DescriptorPath` exists is that it allows programs to correctly handle arbitrary nesting of descriptors.

*However* I just re-read the original description of the PR and realised that passing descriptors makes no sense -- you are only calling `policy(desc)` if there is no descriptor in the black set (in other words, on blobs that have no descriptor pointing to them), so the descriptor will always be an empty struct. The new tests didn't catch this, so you'll need to update the tests so they can exercise this code properly.

oci/casext/gc.go
```diff
@@ -74,11 +81,26 @@ func (e Engine) GC(ctx context.Context) error {
 	}
 
 	n := 0
+sweep:
 	for _, digest := range blobs {
-		if _, ok := black[digest]; ok {
+		desc, ok := black[digest]
+		if ok {
```
cyphar, line 87: This means that `desc` will be an empty `ispec.Descriptor` ...

oci/casext/gc.go
```diff
@@ -74,11 +81,26 @@ func (e Engine) GC(ctx context.Context) error {
...
 			// Digest is in the black set.
 			continue
 		}
+
+		// Iterate over all policy funcs and if any one of them indicates that GC
+		// must not be performed, skip that blob
+		for _, policy := range policies {
+			ok, err := policy(desc)
```
cyphar, line 95: ... when you call `policy` here. So the policy will always see `ispec.Descriptor{}`.

oci/casext/gc_test.go
```diff
@@ -175,3 +176,144 @@ func TestGCWithNonEmptyIndex(t *testing.T) {
...
+	return true, nil
+}
+
+func gcSkipFunc(desc ispec.Descriptor) (bool, error) {
+	return false, nil
+}
+
+func errFunc(desc ispec.Descriptor) (bool, error) {
+	return false, fmt.Errorf("err policy")
+}
```
cyphar, line 190: Ideally these should be functions that actually check that the `Digest` passed is correct. A simple way to do this would be:

```go
func gcOkFunc(t *testing.T, expectedDigest digest.Digest) GCPolicy {
	return func(digest digest.Digest) (bool, error) {
		if digest != expectedDigest {
			t.Errorf("got incorrect digest to gc policy callback: expected %v, got %v", expectedDigest, digest)
		}
		return true, nil
	}
}
```

oci/casext/gc_test.go
```diff
@@ -175,3 +176,144 @@ func TestGCWithNonEmptyIndex(t *testing.T) {
...
+				Size:      size,
+			},
+		},
+	}
+	data, err := json.Marshal(&m)
+	if err != nil {
+		t.Fatalf("error marshaling json: %+v", err)
+	}
+	mr := bytes.NewReader(data)
+	digest, size, err = engine.PutBlob(ctx, mr)
```
cyphar, line 263: `PutBlobJSON` is a simpler way of doing this. You can actually just do

```
digest, size, err := engineExt.PutBlobJSON(ispec.Manifest{
	Versioned: imeta.Versioned{
		SchemaVersion: 2,
	},
	Config: ispec.Descriptor{
		MediaType: ispec.MediaTypeImageIndex,
		Digest:    digest,
		Size:      size,
	},
	Layers: []ispec.Descriptor{
		{
			MediaType: ispec.MediaTypeImageIndex,
			Digest:    digest,
			Size:      size,
		},
	},
})
```

tych0:
> This is unfortunate, since it means the policy function doesn't know the type of the blob, but that is the nature of the problem.

Ugh, yeah. That is unfortunate. Makes total sense though.

rchincha:
Updated the PR with suggested changes.

rchincha:
https://travis-ci.org/github/opencontainers/umoci/jobs/704404847
Is this failure expected?
```loadinternal: cannot find runtime/cgo```

**Comment by cyphar** [comment]
@rchincha That's not the cause of the failure (it's some weird warning we get when running under Travis that I haven't had a chance to debug). The cause of the failure was that the Open Build Service had some downtime in the past few days, so the package installations failed. I've re-triggered the CI run.

File: oci/casext/gc_test.go (hunk 10 above)
Comment (cyphar) [reply oci/casext/gc_test.go:271]: This still needs to be done.

File: oci/casext/gc_test.go [hunk 11]
```diff
@@ -175,3 +177,151 @@ func TestGCWithNonEmptyIndex(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("GC failed: %+v", err)
+	}
+
+	b, err = engine.ListBlobs(ctx)
+	if err != nil {
+		t.Fatalf("unable to list blobs: %+v", err)
+	}
+	if len(b) != 2 {
+		t.Fatalf("expected single-entry blob list after GC")
```
cyphar, line 325: Error message needs to be changed since we now expect two blobs to survive.

oci/casext/gc_test.go
```diff
@@ -175,3 +177,151 @@ func TestGCWithNonEmptyIndex(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("GC failed: %+v", err)
+	}
+	b, err := engine.ListBlobs(ctx)
+	if err != nil {
+		t.Fatalf("unable to list blobs: %+v", err)
+	}
+	if len(b) == 1 {
+		t.Fatalf("didn't expect single-entry blob list after skip GC policy")
+	}
```
cyphar, line 312: This should check `if len(b) != 3` (`== 1` should've been `== 2` but `!= 3` is less fragile and more accurately describes the behaviour we're testing) and have an updated error message.

cyphar:
LGTM.

tych0:
LGTM.

PR #341, opened 2020-07-03
*: improve test coverage

cyphar:
We weren't correctly accounting for the coverage of invocations which failed (non-zero exit code), but in addition there are some obvious error paths we weren't bothering to test (and some error paths were missing). So add some trivial tests and clean up those error paths.


cyphar:
LGTM.

/cc @tych0 @vbatts

tych0:
LGTM,

> In future we should probably get rid of VERSION and just store the
> version inside the Go source code so that "go get" gives you a valid
> version.

I came across this recently in a different project; is go get sill something people care about supporting? I don't, but maybe I should...

PR #343, opened 2020-07-11
ci: rework Travis setup and migrate to codecov

cyphar:
We can cache the umoci/ci image using Travis's built-in caching support,
as well as reduce duplication between our test runs by moving the unit
and validation tests to separate stages.

We will handle code coverage reports through CodeCov rather than
manually making use of the collation features (though we will still
allow people to run an equivalent of the CI locally with "make ci").


cyphar:
This is still a draft because (annoyingly) Codecov counts code coverage differently to our existing `hack/ci-coverage.sh` script and by switching to Codecov our coverage went down to ~77.5% which is less than our normal cut-off of 80%...

cyphar:
LGTM. We can increase CI coverage to over 80% in future PRs.

cyphar:
LGTM. We can increase CI coverage to over 80% in future PRs.

/cc @tych0 @vbatts

tych0:
LGTM, other than a comment which is that I don't think that 100% coverage is really a sane goal :)

cyphar:
I wouldn't aim for 100% either. There are a few "easy" things we can do to increase code coverage (there are a few sanity checks that we do both in `cmd/umoci` and `oci/` -- so it wouldn't hurt to write a few unit tests verifying that they work properly). Also some proper unit tests for `oci/cas{,ext}` wouldn't hurt either.

PR #344, opened 2020-07-15
mutate: Add() should tell you what it added

tych0:
a couple of commits to allow introspection of mutator

### DIFF
--- a/mutate/mutate.go
+++ b/mutate/mutate.go
@@ -155,6 +155,17 @@ func (m *Mutator) Config(ctx context.Context) (ispec.ImageConfig, error) {
 	return m.config.Config, nil
 }
 
+// Manifest returns the current (cached) image manifest. This is what will be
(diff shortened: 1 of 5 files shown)


mutate/mutate.go
```diff
@@ -280,24 +291,26 @@ func (m *Mutator) add(ctx context.Context, reader io.Reader, history *ispec.Hist
...
 		Digest:    digest,
 		Size:      size,
-	})
-	return nil
+	}
+	m.manifest.Layers = append(m.manifest.Layers, desc)
+	return desc, nil
 }
 
 // AddNonDistributable is the same as Add, except it adds a non-distributable
```
cyphar, line 316: `AddNonDistributable` probably should get the same treatment.

tych0:
Ping?

cyphar:
LGTM, sorry for not looking at this for a while. :cry:

PR #347, opened 2020-08-15
-buildmode=pie is not supported for some arch

XiaodongLoong:
-buildmode=pie is not supported for some arch
error logs:
-buildmode=pie not supported on linux/mips64le
reference: https://github.com/containerd/containerd/blob/master/Makefile.linux#L23

use GO111MODULE=on and mod=vendor flag to build umoci
build timeout
reference: https://github.com/containers/skopeo/blob/master/Makefile#L37

Signed-off-by: Xiaodong Liu <liuxiaodong@loongson.cn>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,18 @@ SHELL = /bin/bash
 # Go tools.
 GO ?= go
 GO_MD2MAN ?= go-md2man
-export GO111MODULE=on
+ifeq ($(shell go help mod >/dev/null 2>&1 && echo true), true)
+  export GO111MODULE=on
+  MOD_VENDOR=-mod=vendor
+endif
+GOOS=$(shell go env GOOS)
+GOARCH=$(shell go env GOARCH)
+GO_BUILDMODE="-buildmode=pie"
+ifeq ($(GOOS), linux)
+	ifeq ($(GOARCH), $(filter $(GOARCH),mips mipsle mips64 mips64le ppc64 riscv64))
+		GO_BUILDMODE=
+	endif
+endif
 
 # Set up the ... lovely ... GOPATH hacks.
 PROJECT := github.com/opencontainers/umoci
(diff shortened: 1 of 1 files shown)


XiaodongLoong:
There is a build error:
$ make GO111MODULE=off GOARCH=386 local-validate-build
444WARNING: This make invocation will fetch and run code from https://codecov.io/.
445go build -mod=vendor  -tags "" "-buildmode=pie" -ldflags "-s -w -X github.com/opencontainers/umoci.gitCommit="5caa73e2aa610ed220475aaf76bd1be704cbb32b" -X github.com/opencontainers/umoci.version=0.4.6+dev" -o /dev/null github.com/opencontainers/umoci/cmd/umoci
446build flag -mod=vendor only valid when using modules
447Makefile:174: recipe for target 'local-validate-build' failed
448make: *** [local-validate-build] Error 1
449The command "make GO111MODULE=off GOARCH=386 local-validate-build" exited with 2.
How to fix,
I add a new variable LOCAL_VALIDATE_BUILD_FLAGS, that has no `-mod=vendor`

XiaodongLoong:
@cyphar PTAL, thanks!

Makefile
```diff
@@ -68,8 +79,9 @@ BASE_FLAGS := ${BUILD_FLAGS} -tags "${BUILDTAGS}"
 BASE_LDFLAGS := -s -w -X ${PROJECT}.gitCommit=${COMMIT} -X ${PROJECT}.version=${VERSION}
 
 # Specific build flags for build type.
-DYN_BUILD_FLAGS := ${BASE_FLAGS} -buildmode=pie -ldflags "${BASE_LDFLAGS}"
-TEST_BUILD_FLAGS := ${BASE_FLAGS} -buildmode=pie -ldflags "${BASE_LDFLAGS} -X ${PROJECT}/pkg/testutils.binaryType=test"
+DYN_BUILD_FLAGS := ${MOD_VENDOR} ${BASE_FLAGS} ${GO_BUILDMODE} -ldflags "${BASE_LDFLAGS}"
+LOCAL_VALIDATE_BUILD_FLAGS := ${BASE_FLAGS} ${GO_BUILDMODE} -ldflags "${BASE_LDFLAGS}"
```
cyphar, line 83: I don't think this is necessary -- have you checked whether removing `GO111MODULE=off` in `.travis.yml` fixes the problem? The reason we originally added it was to work around some weird build errors on older Go versions which might've been fixed by now.

PR #348, opened 2020-09-28
mutate: allow for custom compression

tych0:
at least via the API for now :)

### DIFF
--- a/mutate/compress.go
+++ b/mutate/compress.go
@@ -1,0 +1,69 @@
+package mutate
+
+import (
+	"io"
+	"io/ioutil"
+	"runtime"
+
+	gzip "github.com/klauspost/pgzip"
+	"github.com/pkg/errors"
+)
+
+// Compressor is an interface which users can use to implement different
+// compression types.
+type Compressor interface {
+	// Compress sets up the streaming compressor for this compression type.
+	Compress(io.Reader) (io.ReadCloser, error)
+
+	// MediaTypeSuffix returns the suffix to be added to the layer to
+	// indicate what compression type is used, e.g. "gzip", or "" for no
+	// compression.
+	MediaTypeSuffix() string
+}
+
+type noopCompressor struct{}
+
+func (nc noopCompressor) Compress(r io.Reader) (io.ReadCloser, error) {
+	return ioutil.NopCloser(r), nil
+}
+
+func (nc noopCompressor) MediaTypeSuffix() string {
+	return ""
+}
+
+// NoopCompressor provides no compression.
+var NoopCompressor Compressor = noopCompressor{}
+
+// GzipCompressor provides gzip compression.
+var GzipCompressor Compressor = gzipCompressor{}
+
+type gzipCompressor struct{}
+
+func (gz gzipCompressor) Compress(reader io.Reader) (io.ReadCloser, error) {
+	pipeReader, pipeWriter := io.Pipe()
+
+	gzw := gzip.NewWriter(pipeWriter)
+	if err := gzw.SetConcurrency(256<<10, 2*runtime.NumCPU()); err != nil {
+		return nil, errors.Wrapf(err, "set concurrency level to %v blocks", 2*runtime.NumCPU())
+	}
+	go func() {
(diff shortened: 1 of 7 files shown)


mutate/compress.go
```diff
@@ -0,0 +1,76 @@
+package mutate
```
cyphar, line 1: All of the exported types in here need go docs.

mutate/compress.go
```diff
@@ -0,0 +1,76 @@
...
+	r io.Reader
+}
+
+func (wc withClose) Read(p []byte) (n int, err error) {
+	return wc.r.Read(p)
+}
+
+func (wc withClose) Close() error {
+	return nil
+}
```
cyphar, line 27: You can switch to `ioutil.NopCloser` and remove this wrapper.

mutate/compress.go
```diff
@@ -0,0 +1,76 @@
...
+			// #nosec G104
+			_ = pipeWriter.CloseWithError(errors.Wrap(err, "close pipe writer"))
+		}
+	}()
+
+	return pipeReader, nil
+}
+
+func (gz gzipCompressor) MediaType() string {
+	return ispec.MediaTypeImageLayerGzip
```
cyphar, line 75: Technically this isn't quite right (you could gzip compress things other than tar archives) -- this should return the passed media type with `+gzip` appended. Then again, I have a feeling that all of this should actually be handled by returning the mediatype suffix (`"gzip"`, `"ztsd"`, `""`, etc) for the compression algorithm and then the caller can add the suffix themselves.

There have been discussions about switching aware from string-const based media-types (and instead actually parse them so combined map functions like encryption+compression can be represented properly). I don't think that's going to happen soon, but I think representing compression functions as just adding to an existing media-type rather than having their own media-type is a bit nicer.

mutate/mutate.go
```diff
@@ -291,21 +272,20 @@ func (m *Mutator) add(ctx context.Context, reader io.Reader, history *ispec.Hist
...
+	digest, size, err := m.add(ctx, r, history, compressor)
 	if err != nil {
 		return desc, errors.Wrap(err, "add layer")
 	}
 
 	// Append to layers.
 	desc = ispec.Descriptor{
-		// TODO: Detect whether the layer is gzip'd or not...
-		MediaType: ispec.MediaTypeImageLayerGzip,
+		MediaType: compressor.MediaType(),
```
cyphar, line 288: As I said in the other comment this should be something like *shudder* `ispec.MediaTypeImageLayer + compressor.MediaType()`. I get that you folks don't use layers `.tar.gz` layers -- but that means we should extend `Mutator` to separately support using a custom media-type IMHO. Custom compression is kind of separate to custom layer media-types.

cyphar (changes requested):
Seems reasonable, just a few API questions and cleanups that I reckon would be a good idea.

cyphar:
And yeah, I can cook up the CLI UX for it.

mutate/compress.go
tych0: Ok, I refactored this a bit to take this into account. Let me know how it looks now.

tych0:
408 while uploading code coverage seems like a transient failure... I re-ran travis.

mutate/mutate.go
```diff
@@ -291,45 +272,25 @@ func (m *Mutator) add(ctx context.Context, reader io.Reader, history *ispec.Hist
...
-		Digest:    digest,
-		Size:      size,
-	}
-	m.manifest.Layers = append(m.manifest.Layers, desc)
-	return desc, nil
-}
-
-// AddNonDistributable is the same as Add, except it adds a non-distributable
-// layer to the image.
-func (m *Mutator) AddNonDistributable(ctx context.Context, r io.Reader, history *ispec.History) (ispec.Descriptor, error) {
```
cyphar, line 318: Huh, I didn't consider that we could drop `AddNonDistributable` entirely. Nice. :smile_cat:

cyphar (approved):
LGTM. The number of arguments to `mutator.Add` is getting a bit long but we can always deal with that some other time.

PR #350, opened 2020-11-21
Added support for zstd compression

### LINKED ISSUE #316: Add support for zstd compression
Per https://github.com/opencontainers/image-spec/issues/803 using zstd compression seems to be on peoples minds.

After reading about it, it seems it's been designed with parallelism in mind, among just being faster on a single core as well. I think it would definitely speed up some of our use cases, at least.

nishakm:
Using github.com/klauspost/compress go module has brought in a
list of other compression implementations. Therefore the vendor
folder has gotten larger. Some files were also modified due to
the bump up in compress's version.

Apart from these, this commit adds an implementation for zstd
compression and a corresponding test.

Fixes #316

Signed-off-by: Nisha K <nishak@vmware.com>


nishakm:
@tych0 I'm not sure why the integration tests are failing 😅. Any hints on what to do?

tych0:
Looks like something is busted in the umoci CI docker image, probably a git master version of bats or something. It's failing on umoci master the same way:

https://travis-ci.org/github/opencontainers/umoci/jobs/746582491

@cyphar how are these umoci CI images built and uploaded?

cyphar:
The CI images are built during the CI and are then cached using Travis, but due to some Docker limitations the only way you can do this nicely is to use `--cache-from`. I did originally seed the image with a locally-built one, but in principle you shouldn't end up with them being used by actual tests. I might need to rethink this, but I'll push an updated version now.

cyphar:
But to actually build the CI image I just run `make ci-image` and `docker push umoci/ci`. I can give you access to that organisation on Docker Hub if you like.

cyphar:
I think the CI issue is related to the `1.2.1` update and appears to be a bats regression of some sort. Still looking into it...

cyphar:
This was a regression but was fixed in bats-core/bats@9b2b659e5ba76. I'll backport it to openSUSE then re-run CI.

cyphar:
Alright, now the CI has finally been wrangled I can review this (sorry about that). The main thing missing is that umoci cannot decompress zstd at the moment, and ideally I'd like it if we added support for that too (as well as a CLI UX for it). But I can work on that separately.

LGTM.

cyphar:
@tych0 LGTM from you as well?

tych0:
Yep, sorry. LGTM as well.

PR #354, opened 2021-01-25
layer: ignore overlay.impure xattr

tych0:
See comment for details; basically this is a during-image-build-time-only
thing.


tych0:
I suppose we should probably ignore *all* the overlay xattrs (e.g. opaque or other things) with similar kinds of semantics... perhaps I should switch this patch to just ignore all of them. Let me know what you think.

cyphar:
Yeah it might not make sense to include any `trusted.overlay` attributes, though I would be worried a future attribute being added which actually should be kept by umoci...

tych0:
Yeah, perhaps I'll just update this and add all of the existing ones now, I think there's only ~6.

tych0:
Ping?

cyphar (approved):
LGTM, sorry.

PR #359, opened 2021-02-08
Add 'umoci copy' subcommand

### LINKED ISSUE #345: support pulling images with OCI Distribution Spec API
@cyphar Any reason not to have support for pulling OCI images directly without skopeo?

e.g. 
```
$ skopeo copy docker://alpine:latest oci:/tmp/alpine:latest
$ umoci unpack --image /tmp/alpine /tmp/alpine-bundle
```
could be written as 
```
$ umoci unpack --image docker://alpine:latest /tmp/alpine-bundle
```

jdolitsky:
- Support for pulling oci:// URL into local store
- Support for pushing from local store to registry
- Downloads blobs only when not found in local store
- Currently only supports OCI v1 manifest media type
- Does not support docker:// URLs or Docker manifests
- Does not yet support chunked blob uploading

Resolves #345


cmd/umoci/copy-remote-to-local.go
```diff
@@ -0,0 +1,217 @@
...
+		log.Debugf("actual manifest digest matches expected digest (%s)", expectedContentDigest)
+
+		// Note: only "application/vnd.oci.image.manifest.v1+json" supported for now
+		mediaType := resp.Header().Get(manifestMediaTypeHeader)
+		if mediaType != v1.MediaTypeImageManifest {
+			return errors.New(
+				fmt.Sprintf("Content-Type header for image manifest invalid: %s", mediaType))
+		}
+
+		d, s, err := engine.PutBlob(engineContext, bytes.NewReader(resp.Body()))
```
tych0, line 108: As far as I can tell, resp.Body() here is: https://pkg.go.dev/github.com/go-resty/resty/v2#Response.Body which will end up reading the whole layer into memory. Seems like you want RawBody() here?

https://pkg.go.dev/github.com/go-resty/resty/v2#Response.RawBody

tych0: Oh, derp. I meant to put this on the layer one below. For this one I think it's unavoidable: you definitely want to read the manifest bytes into memory.

tych0:
One small nit on this, but otherwise it looks like it might be useful to us. How close do you feel this is to merging?

cyphar:
@tych0 

> One small nit on this, but otherwise it looks like it might be useful to us. How close do you feel this is to merging?

I've been chatting to @jdolitsky about this for a little bit -- my current plan is for me to rework this PR so that it's implemented as a `cas.Engine` backend once I get back from vacation next week.

tych0:
> my current plan is for me to rework this PR so that it's implemented as a cas.Engine backend once I get back from vacation next week.

Excellent, thanks!

tych0:
What's the status of this? I'd like to get some features into containers/image soon-ish, but it would be better to have them here, if possible :)

cyphar:
I haven't touched this in a while. I can try to knock up a PoC using the CAS API -- what exactly did you want to get into containers/image? Switching to `umoci copy` would probably be somewhat non-trivial even if I did get it working soon.

tych0:
Mostly I don't want it to compress things automatically that aren't gzip :)

cyphar:
I feel the the only slight complication is going to be chunked uploads and downloads. I started working on a PoC of this, it shouldn't be *that* bad on paper but implementing a generic `umoci copy` API might be slightly tricky and I think the `CasEngineExt` API needs to be reworked somewhat in order to be able to handle arbitrary CAS implementations to wrap).

PR #360, opened 2021-02-18
oci: unpack: slurp up raw layer stream before Close()

### LINKED ISSUE #358: Question: unpack: reader size mismatch discarding trailing archive bits
This is a question rather than an issue report, as I cannot reproduce with any other image, and our reporter cannot provide the image they have which causes this behavior. Please feel free to close for lack of information!

We have had a report of an umoci unpack failing on an image (which unfortunately the reporter cannot provide). 

https://github.com/hpcng/singularity/issues/5799

Umoci reports:

```
discard trailing archive bits: expected 31101 bytes (only 31100 bytes in stream): verified reader size mismatch
```

Looking at the code, umoci is discarding padding etc. here as expected but I'm not clear why the stream could end up 1 byte shorter than expected.

Just wondering whether this has ever been seen before?

cyphar:
It's possible there's some trailing bytes after the gzip stream (which
truth be told, probably shouldn't be there), so in order to avoid
spurious errors on layer unpacking with slightly-broken images, just
slurp up that data.

But we should output logging information since this indicates that the
original image builder probably has a bug of some kind.

We also add code to slurp trailing bytes in VerifiedReadCloser.Close()
This is to ensure that we handle both good and bad cases sanely --
namely, that we always are digesting the *full* blob when Close() is
called. This avoids us producing intermittent errors if blobs have extra
trailing bytes (such as garbage at the end of a gzip stream), as well as
hardening us against incorrect digests (though it should be noted this
case cannot be triggered without disabling size verification -- the size
verficiation we have already blocks this issue).

Fixes #358

### DIFF
--- a/oci/casext/walk.go
+++ b/oci/casext/walk.go
@@ -101,7 +101,7 @@ var ErrSkipDescriptor = errors.New("[internal] do not recurse into descriptor")
 //       more than once. This is quite important for remote CAS implementations.
 type WalkFunc func(descriptorPath DescriptorPath) error
 
-func (ws *walkState) recurse(ctx context.Context, descriptorPath DescriptorPath) error {
+func (ws *walkState) recurse(ctx context.Context, descriptorPath DescriptorPath) (Err error) {
 	log.WithFields(log.Fields{
 		"digest": descriptorPath.Descriptor().Digest,
 	}).Debugf("-> ws.recurse")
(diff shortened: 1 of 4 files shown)


cyphar:
Hmmm, the failure is real but it comes from `io.Copy` returning `io.EOF` which should absolutely never happen. This only happens with the `pgzip` reader, so there's something really weird going on...

cyphar:
We need klauspost/pgzip#40 before we can do this.

tych0:
FWIW, this LGTM whenever all the stuff you need to land lands, feel free to merge.

cyphar:
Thanks, I might just merge this as-is since the pgzip stuff isn't actually necessary now that I've implemented it by just copying from the actual file stream.

tych0:
Sure, sounds good to me.

cyphar:
LGTM.

PR #363, opened 2021-03-12
Fuzzing: Add layer fuzzer

AdamKorcz:
This PR adds a fuzzer that takes the pseudo-random data from the fuzzer and generates a baseDir with files containing parts of the input, which is then passed to `GenerateLayer`.


AdamKorcz:
@cyphar I am having a bit of trouble with passing the CI for this PR. A new dependency is introduced, and `go mod tidy` and `go mod vendor` do not seem to fix this.

cyphar:
I've carried this in #365.

PR #369, opened 2021-04-08
makefile: fix bad build flags

cyphar:
Fix mistake in the Makefile which prevents the version field (as well as
some other build flags) from being passed to "go build".

Fixes: 6fbd32e48b66 ("Make Makefile more portable")


cyphar:
I will also add a quick test for this as well...

cyphar:
LGTM.

PR #370, opened 2021-04-19
ci: verify codecov hashes

cyphar:
We don't have any secrets in our CI, but it doesn't hurt to be sure now
that they actually publish hashes in a way that we can easily verify. We
can also pin the version (though in the future we should probably pull
the script from codecov.io and verify it using GitHub separately).


tych0:
heh lgtm :)

PR #371, opened 2021-04-19
Add 2 fuzzers

AdamKorcz:
This PR does the following:

1. Add two fuzzers. One for the layer package and one for the mutate package.
2. Replace the AdamKorcz/go-fuzz-headers with the AdaLogics/go-fuzz-headers dependency.
3. Add a few comments to the existing layer fuzzer.


cyphar:
Is it possible to make the `_fuzzer.go` files only be built with a specific build tag (like `fuzzer` or `gofuzz` or something)? The reason I ask is because each new fuzzer reduces our code coverage metrics. :grimacing:

AdamKorcz:
> Is it possible to make the _fuzzer.go files only be built with a specific build tag (like fuzzer or gofuzz or something)? The reason I ask is because each new fuzzer reduces our code coverage metrics. 😬

Yep, I will add it!

AdamKorcz:
@cyphar I added the `// +build gofuzz` tag to the mutate fuzzer. Was that what you had in mind?

mutate/mutate_fuzzer.go
```diff
@@ -0,0 +1,179 @@
+// +build gofuzz
+/*
```
cyphar, line 2: You need to add one more blank space between `// +build` and the other comment. Build comments are super particular about whitespace.

cyphar:
@AdamKorcz 

> @cyphar I added the `// +build gofuzz` tag to the mutate fuzzer. Was that what you had in mind?

Yup, but can you add it to all of the other `_fuzzer.go` files as well?

mutate/mutate_fuzzer.go
```diff
@@ -0,0 +1,179 @@
...
+	imeta "github.com/opencontainers/image-spec/specs-go"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/umoci/oci/cas"
+	casdir "github.com/opencontainers/umoci/oci/cas/dir"
+	"github.com/opencontainers/umoci/oci/casext"
+)
+
+// setup() does the necessary setup for the fuzzer
+// It takes a data parameter provided by the fuzzer
+func setup(dir string, data []byte) (cas.Engine, ispec.Descriptor) {
```
cyphar, line 41: You can't call this function `setup` -- it breaks the integration tests build because there is a function called `setup` in `mutate/mutate_test.go`.

AdamKorcz:
@cyphar 

> Yup, but can you add it to all of the other _fuzzer.go files as well?

I checked the existing fuzzers, they all already have the build comment.

cyphar:
>  I checked the existing fuzzers, they all already have the build comment.

Ah, okay. I missed that. Thanks!

cyphar (approved):
LGTM.

mutate/mutate_fuzzer.go
```diff
@@ -0,0 +1,180 @@
...
+// FuzzMutate implements the fuzzer
+func FuzzMutate(data []byte) int {
+	c := fuzz.NewConsumer(data)
+	byteArray, err := c.GetBytes()
+	if err != nil {
+		return -1
+	}
+	dir, err := ioutil.TempDir("", "umoci-TestMutateAdd")
+	if err != nil {
+		return -1
```
tych0, line 133: Maybe I missed it, but what do -1, 0, 1, and panic() signify here? Seems like these -1s should be panic()s maybe, and 0 and 1 are fuzzer failure/success codes?

cyphar: From my understanding there are 4 possible results you get from a fuzzed function:

 * If it panics, that indicates a crash (bug) and will be kept as a "bad" sample.
 * If it returns `-1`, there was some other error than an input-related crash.
 * If it returns `0`, the input was invalid.
 * If it returns `1`,  the input was valid.

You don't want to `panic()` from a `Fuzz...` function because that means any input which happens to coincide with an unrelated error will be treated as a "bad" sample.

AdamKorcz: Regarding the return values, the relevant documentation can be found here: https://github.com/dvyukov/go-fuzz#usage

However, [the significance of these is low](https://github.com/tailscale/tailscale/pull/1115#discussion_r556171316) and using different return values is more a matter of code structure and formatting than performance, ie: returning `-1` during set up and initialization (`if err!=nil`), returning `0` when fuzzing the target (`if err!=nil`), and returning `1` if the fuzzers succeeds.

The fuzzing engine does not rely on the return values to determine if the input was valid.

Regarding the panics: You don't want panics in a fuzzer, if these are fuzz blockers that prevent it from running continuously. In our case, we build the fuzzers in libfuzzer mode, and panics will stop the fuzzer, so they will be fuzz blockers. However, I would prefer to have the panics in the fuzzer for now, as it will be good to know if the fuzzer fails during initialization, and panics will provide this feedback whereas returning `-1` will not.

tych0: > 

Regarding the panics: You don't want panics in a fuzzer, if these are fuzz blockers that prevent it from running continuously. In our case, we build the fuzzers in libfuzzer mode, and panics will stop the fuzzer, so they will be fuzz blockers. However, I would prefer to have the panics in the fuzzer for now, as it will be good to know if the fuzzer fails during initialization

But isn't the goal to find panic()s during initialization? For example, you want to find invalid array accesses from PutBlobJSON() or whatever; panic()ing on an error from that is confusing to me. Seems like it should be a `return 0` as well? Or is the goal to move it to `return 0` in the near future? (If so, can you put that in the commit message?)

AdamKorcz: > is the goal to move it to return 0 in the near future?

This was the plan, but I have changed all panics to return 0's now. I am marking this one as resolved, but feel free to reopen if you think otherwise.

tych0:
LGTM

AdamKorcz:
Cool. Squash commits and we can merge?

tych0:
Yep, if you want to squash to something sensible I'm happy to merge.

tych0:
Failure was, `curl: (22) The requested URL returned error: 408` from coverage, let's see if a re-run fixes it...

AdamKorcz:
Sounds good, but can I just get a small commit in before merging?

AdamKorcz:
Alright, there were a few `panic`'s in the unpack_fuzzer we could get rid of. All ready from my side!

tych0:
Thanks!

PR #372, opened 2021-04-19
Add documentation for running fuzzers

AdamKorcz:
This PR adds documentation on running the fuzzers locally. 

It adds the `/test/fuzzing` that holds the Dockerfile and the Readme.

### DIFF
--- a/Dockerfile.fuzz
+++ b/Dockerfile.fuzz
@@ -1,0 +1,17 @@
+FROM golang:1.14
+
+RUN apt-get update && apt-get install -y clang
+
+RUN go get -u github.com/dvyukov/go-fuzz/go-fuzz \
+	github.com/dvyukov/go-fuzz/go-fuzz-dep \
+	github.com/dvyukov/go-fuzz/go-fuzz-build
+
+COPY . $GOPATH/src/github.com/opencontainers/umoci
+
+WORKDIR $GOPATH/src/github.com/opencontainers/umoci
+
+RUN go clean --modcache && \
+    go mod tidy && \
+    go mod vendor && \
+    rm -r ./vendor
+
--- a/test/fuzzing/README.md
+++ b/test/fuzzing/README.md
@@ -1,0 +1,30 @@
+# Fuzzing umoci
+
+Umoci has a series of fuzz tests. These are implemented by way of [go-fuzz](https://github.com/dvyukov/go-fuzz).
+
+## Running the fuzzers
+
+To run the fuzzers, first build the fuzzer image from the root of this repository:
+
+```bash
+sudo docker build -t umoci-fuzz -f Dockerfile.fuzz .
+```
+Next, get a shell in the container:
+```bash
+sudo docker run -it umoci-fuzz
+```
+At this point, you can navigate to any directory that has a fuzzer and build it:
+
+```bash
+cd $PATH_TO_FUZZER
+go-fuzz-build -libfuzzer -func=FUZZ_NAME && \
+clang -fsanitize=fuzzer PACKAGE_NAME.a -o fuzzer
+```
+`FUZZ_NAME` will typically be `Fuzz`, but in some cases the respective fuzzers will have more descriptive names. 
+
+If you encounter any errors when linking with `PACKAGE_NAME.a`, simply `ls` after running `go-fuzz-build...`, and you will see the archive to link with.
+
+If everything goes well until this point, you can run the fuzzer:
+```bash
+./fuzzer
+```


test/fuzzing/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM golang:1.14
+
+WORKDIR /go
+
+RUN apt-get update && apt-get install -y clang
+
+RUN git clone https://github.com/opencontainers/umoci \
+    $GOPATH/src/github.com/opencontainers/umoci && \
+    cd $GOPATH/src/github.com/opencontainers/umoci && \
```
cyphar, line 9: If you move the dockerfile to the root of the repo (and call it `Dockerfile.fuzz` or something) then you don't need to do this clone step -- which would be ideal because then you could run the fuzzer on a locally-modified copy or branch.

test/fuzzing/README.md
```diff
@@ -0,0 +1,30 @@
...
+Next, get a shell in the container:
+```bash
+sudo docker run -it umoci-fuzz
+```
+At this point, you can navigate to any directory that has a fuzzer and build it:
+
+```bash
+cd $PATH_TO_FUZZER
+go-fuzz-build -libfuzzer -func=FUZZ_NAME && \
+clang -fsanitize=fuzzer PACKAGE_NAME.a -o fuzzer
```
cyphar, line 21: Is it possible to make this a script which is run on `docker build` so that users don't need to find out where the fuzzers live.

AdamKorcz: Yes, I have an idea to do that. Might I suggest we do it in a separate PR, as it might require a few modifications to the existing fuzzers.

cyphar: Sounds good, thanks!

test/fuzzing/Dockerfile
AdamKorcz: I just pushed an update. Does that look better to you? If so, I will just amend the documentation.

Dockerfile.fuzz
```diff
@@ -0,0 +1,16 @@
...
+
+RUN cd $GOPATH/src/github.com/opencontainers/umoci && \
+    go clean --modcache && \
+    go mod tidy && \
+    go mod vendor && \
+    rm -r ./vendor
+
+RUN go get -u github.com/dvyukov/go-fuzz/go-fuzz \
+	github.com/dvyukov/go-fuzz/go-fuzz-dep \
+	github.com/dvyukov/go-fuzz/go-fuzz-build
```
cyphar, line 14: Can you move this to before the `COPY` -- that way it doesn't get run each time the source directory changes.

Dockerfile.fuzz
```diff
@@ -0,0 +1,16 @@
...
+
+WORKDIR $GOPATH/src/github.com/opencontainers/umoci
```
cyphar, line 16: If you put this before the `RUN go clean ...` you don't need to do `cd` in that command.

test/fuzzing/Dockerfile
cyphar: Yup, that looks better (with some minor nits about the ordering of instructions).

AdamKorcz:
@cyphar 
All things should have been addressed. Let me know what you think

Dockerfile.fuzz
cyphar: I think this needs to go above line 11 in order to work properly?

cyphar (commented):
Aside from one final nit, LGTM.

test/fuzzing/README.md
```diff
@@ -0,0 +1,30 @@
+# Fuzzing umoci
+
+Umoci has a series of fuzz tests. These are implemented by way of [go-fuzz](https://github.com/dvyukov/go-fuzz).
+
+## Running the fuzzers
+
+To run the fuzzers, first build a container from the Docker image in `umoci/test/fuzzing`:
+
+```bash
+sudo docker build -t umoci-fuzz .
```
cyphar, line 10: Needs an `-f Dockerfile.fuzz` now.

test/fuzzing/README.md
```diff
@@ -0,0 +1,30 @@
+# Fuzzing umoci
+
+Umoci has a series of fuzz tests. These are implemented by way of [go-fuzz](https://github.com/dvyukov/go-fuzz).
+
+## Running the fuzzers
+
+To run the fuzzers, first build a container from the Docker image in `umoci/test/fuzzing`:
```
cyphar, line 7:
```suggestion
To run the fuzzers, first build the fuzzer image from the root of this repository:
```

test/fuzzing/README.md
cyphar, line 10:
```suggestion
sudo docker build -t umoci-fuzz -f Dockerfile.fuzz .
```

Dockerfile.fuzz
AdamKorcz: It should be there now

cyphar (approved):
LGTM, thanks.

tych0:
LGTM.

Issue #373, opened 2021-07-27
performance regression: GC() (or more generally, Walk()) now reads every blob off of disk and hashes it (twice)


tych0:
As of 4faf753956f69243f72eab667cc5ee0fd74c1d56 we now read every byte of every blob that's opened. Since the Walk() API uses FromDescriptor() to open a blob, every blob is opened during a GC, including data blobs. With the above patch, we end up hashing every byte in the image.

Further, FromDescriptor() uses GetVerifiedBlob(), which uses GetBlob(), both of which use a VerifiedReadCloser, so we actually end up checking each blob's hash *twice* during this verification.

This is a fairly severe performance regression from before, where we only verified blobs that were actually read in their entirety, IIUC.

Perhaps we can implement a .FSCK() which verifies blobs, and leave GC() to just do GC? Not clear how to do this given the current API that opens everything.

Stack trace for posterity:

```
goroutine 1 [runnable]:                                                                                   
github.com/pkg/errors.Cause(0x0, 0x0, 0x2000, 0x2000)                                                                                                                                                               
        /home/tycho/packages/go/pkg/mod/github.com/pkg/errors@v0.9.1/errors.go:275 +0x9e                  
github.com/opencontainers/umoci/pkg/hardening.(*VerifiedReadCloser).Read(0xc0002eab40, 0xc000308000, 0x2000, 0x2000, 0x2000, 0x0, 0x0)                                                                              
        /home/tycho/packages/go/pkg/mod/github.com/opencontainers/umoci@v0.4.7/pkg/hardening/verified_reader.go:157 +0x185                                                                                          
github.com/opencontainers/umoci/pkg/hardening.(*VerifiedReadCloser).Read(0xc0002eab80, 0xc000308000, 0x2000, 0x2000, 0x2000, 0x0, 0x0)                                                                              
        /home/tycho/packages/go/pkg/mod/github.com/opencontainers/umoci@v0.4.7/pkg/hardening/verified_reader.go:127 +0xaa                                                                                           
io.discard.ReadFrom(0x1bd2200, 0xc0002eab80, 0x1130e40, 0xc000450401, 0x7fbf800209e0)                     
        /usr/local/go/src/io/io.go:598 +0x95                                                              
io.copyBuffer(0x1bd4b40, 0x22abb00, 0x1bd2200, 0xc0002eab80, 0x0, 0x0, 0x0, 0xc0002eab80, 0x16, 0xc00099bfe8)                                                                                                       
[...]

PR #374, opened 2021-08-02
mutate: implement an AddExisting() API

tych0:
in order to add layers that may already exist in the image to a mutator, we
would previously have to read the layer off the disk, decompress it, and
recompress it (arriving at the same results, modulo gzip stream weirdness).

Let's instead just implement an API for letting us do this directly,
without all the i/o and CPU usage.

I somehow can't run the CI locally:

```
#6 ERROR: executor failed running [/bin/sh -c zypper mr -d repo-non-oss repo-update-non-oss && 	zypper ar -f -p 5 -g obs://home:cyphar:bats obs-bats && 	zypper ar -f -p 10 -g obs://Virtualization:containers obs-vc && 	zypper ar -f -p 10 -g obs://devel:tools obs-tools && 	zypper ar -f -p 10 -g obs://devel:languages:go obs-go && 	zypper --gpg-auto-import-keys -n ref && 	zypper -n up]: exit code: 4
------
 > [2/7] RUN zypper mr -d repo-non-oss repo-update-non-oss && 	zypper ar -f -p 5 -g obs://home:cyphar:bats obs-bats && 	zypper ar -f -p 10 -g obs://Virtualization:containers obs-vc && 	zypper ar -f -p 10 -g obs://devel:tools obs-tools && 	zypper ar -f -p 10 -g obs://devel:languages:go obs-go && 	zypper --gpg-auto-import-keys -n ref && 	zypper -n up:
------
executor failed running [/bin/sh -c zypper mr -d repo-non-oss repo-update-non-oss && 	zypper ar -f -p 5 -g obs://home:cyphar:bats obs-bats && 	zypper ar -f -p 10 -g obs://Virtualization:containers obs-vc && 	zypper ar -f -p 10 -g obs://devel:tools obs-tools && 	zypper ar -f -p 10 -g obs://devel:languages:go obs-go && 	zypper --gpg-auto-import-keys -n ref && 	zypper -n up]: exit code: 4
make: *** [Makefile:232: ci-image] Error 1
```

So hopefully this passes :)

### DIFF
--- a/mutate/mutate.go
+++ b/mutate/mutate.go
@@ -227,6 +227,22 @@ func (m *Mutator) Set(ctx context.Context, config ispec.ImageConfig, meta Meta, 
 	return nil
 }
 
+func (m *Mutator) appendToConfig(history *ispec.History, layerDiffID digest.Digest) {
+	m.config.RootFS.DiffIDs = append(m.config.RootFS.DiffIDs, layerDiffID)
+
+	// Append history.
+	if history != nil {
+		history.EmptyLayer = false
+		m.config.History = append(m.config.History, *history)
+	} else {
+		// Some tools get confused if there are layers with no history entry.
+		// Especially if you have later layers have history entries (which will
+		// result in the history entries not matching up and everyone getting
+		// quite confused).
+		log.Warnf("new layer has no history entry -- this will confuse many tools!")
+	}
+}
+
 // add adds the given layer to the CAS, and mutates the configuration to
 // include the diffID. The returned string is the digest of the *compressed*
 // layer (which is compressed by us).
@@ -251,19 +267,7 @@ func (m *Mutator) add(ctx context.Context, reader io.Reader, history *ispec.Hist
 
 	// Add DiffID to configuration.
 	layerDiffID := diffidDigester.Digest()
-	m.config.RootFS.DiffIDs = append(m.config.RootFS.DiffIDs, layerDiffID)
-
-	// Append history.
-	if history != nil {
-		history.EmptyLayer = false
-		m.config.History = append(m.config.History, *history)
-	} else {
-		// Some tools get confused if there are layers with no history entry.
-		// Especially if you have later layers have history entries (which will
-		// result in the history entries not matching up and everyone getting
-		// quite confused).
-		log.Warnf("new layer has no history entry -- this will confuse many tools!")
-	}
+	m.appendToConfig(history, layerDiffID)
 	return layerDigest, layerSize, nil
 }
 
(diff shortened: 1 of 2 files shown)


cyphar:
Yeah I need to rework the CI -- the fact we used a custom image in order to get around issues with Travis's caching is particularly gnarly. I'll see if I can figure something out this week.

mutate/mutate.go
```diff
@@ -298,6 +302,27 @@ func (m *Mutator) Add(ctx context.Context, mediaType string, r io.Reader, histor
...
+	if err := m.cache(ctx); err != nil {
+		return errors.Wrap(err, "getting cache failed")
+	}
+
+	// make sure the blob exists
+	blob, err := m.engine.GetBlob(ctx, desc.Digest)
+	if err != nil {
+		return err
+	}
+	blob.Close()
```
cyphar, line 318: Need to check this error. But in line with #373 we should probably just have a stat-like API so that you don't need to read the whole blob if you're just doing an existence check.

mutate/mutate.go
```diff
@@ -298,6 +302,27 @@ func (m *Mutator) Add(ctx context.Context, mediaType string, r io.Reader, histor
...
+
+	// perhaps we should optionally verify the blob's diffid against its compression?
```
cyphar, line 320: If we're going to have an API that computes the DiffID, it's probably better to just make that API generate teh DiffID and use it rather than comparing against a value the caller provides -- but that'd need to be a separate API to this more "light-weight" version where you just add a layer without reading or writing any layer blobs (once I fix the GetBlob thing.)

mutate/mutate.go
cyphar: Can this wait a bit while I fix #373 so you can build on top of it?

cyphar: #375 has `StatBlob` which is probably want you want to do (actually maybe we don't even need to check if the blob exists?).

mutate/mutate.go
tych0: Yeah, and I really don't want to compute the DiffID since it will be expensive in most cases.

cyphar:
Can you rebase this on top of master? GitHub can't seem to figure out that a new CI config means it should re-run the CI jobs on all open PRs.

mutate/mutate.go
```diff
@@ -298,6 +302,29 @@ func (m *Mutator) Add(ctx context.Context, mediaType string, r io.Reader, histor
...
+	}
+
+	// make sure the blob exists
+	present, err := m.engine.StatBlob(ctx, desc.Digest)
+	if err != nil {
+		return err
+	}
+	if !present {
+		return errors.Errorf("blob %s not present", desc.Digest)
+	}
```
cyphar, line 320: What if we didn't check for the blob's existence? It isn't really necessary for safety reasons, and if the caller really wants to make sure the blob exists they can call StatBlob (or just try to do a walk of the descriptor tree).

tych0: Sure, I'll drop it.

cyphar (approved):
LGTM.

PR #375, opened 2021-08-05
oci: casext: walk: do not attempt to recurse into un-parseable blobs 

cyphar:
Since all of these blobs are not parseable, we won't be able to fetch
sub-descriptors from them, and thus there's no real point trying to
recurse into them. Before commit 4faf753 ("pkg: hardening: slurp
trailing bytes in VerifiedReadCloser.Close()") this wasn't too big of a
deal, but after that commit Close() is quite expensive for blobs you
don't use.

Reported-by: Tycho Andersen <tycho@tycho.pizza>

### DIFF
--- a/oci/cas/cas.go
+++ b/oci/cas/cas.go
@@ -70,7 +70,20 @@ type Engine interface {
 
 	// GetBlob returns a reader for retrieving a blob from the image, which the
 	// caller must Close(). Returns ErrNotExist if the digest is not found.
+	//
+	// This function will return a VerifiedReadCloser, meaning that you must
+	// call Close() and check the error returned from Close() in order to
+	// ensure that the hash of the blob is verified.
+	//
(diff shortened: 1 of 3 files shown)


cyphar:
@tych0 This should fix #373, can you try it out and see?

tych0:
[sorry for the delay, was in the woods for part of last week] Yep, I just tested this and it fixes things for me, thanks. Not sure what you want to do about CI, though. Should I just force-merge?

cyphar:
I'll forcefully fix up the CI this week and we can properly review things afterwards.

cyphar:
Thanks for the ping, I just managed to wrangle the CI into working under Github Actions -- I will need to push the fix directly to master though because you can't really PR-based-review a CI change like this one (it won't work until the CI config is on master).

tych0 (approved):

PR #376, opened 2021-08-30
fix(Makefile): avoid usage of which

CameronNemo:
which is not available in some build environments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # Use bash, so that we can do process substitution.
-SHELL := $(shell which bash)
+SHELL := bash
 
 # Go tools.
 GO ?= go


cyphar:
Sorry can you rebase this on top of master? We just switched CIs and GitHub can't seem to figure out that means it should re-run the CI jobs on all open PRs.

tych0:
Github allows maintainers to force push to forks now by default when people make a PR, so I went ahead and rebased it.

tych0:
Seems like `command` is not available in all build environments either :)

```
 make: -c: Command not found
 ```

**Comment by tych0** [comment]
```
diff --git a/Makefile b/Makefile
index cd3616b..04b0c2c 100644
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # Use bash, so that we can do process substitution.
-SHELL := $(shell command -v bash)
+SHELL := $(shell bash -c 'command -v bash')
 
 # Go tools.
 GO ?= go
```

seems to work for me, perhaps there are better ways though.

CameronNemo:
@tych0 that is... odd. In my testing, dash, ash, bash, oksh, zsh, and even fish support `command -v <foo>`. Are you using csh are something? I am struggling to reproduce the error from the CI locally.

Regardless, I was able to get it working locally with just `SHELL := bash`; that works in CI too. If you would prefer your solution I would not mind.

tych0:
I'm using bash, and I believe CI is too.

cyphar:
I'm not sure why I used `$(shell which bash)` in the first place -- I think Make doesn't mind if it's just `bash`? If it works, it works. :wink:

cyphar (approved):
LGTM.

PR #379, opened 2021-09-17
*: use go1.17-friendly go:build tags

cyphar:
Go 1.17 adds the new "//go:build ..." style of build tags, and gofmt now
requires them to be used so add them -- CI is failing otherwise.


tych0:
I think you can drop the other ones if you change the go version in go.mod, but maybe it's too early to switch everyone to 1.17 wholesale? Anyway, this looks fine to me.

tych0 (approved):

PR #395, opened 2021-09-24
*: use new protobuf package

zhsj:
Replace github.com/golang/protobuf with google.golang.org/protobuf.
The old one is deprecated.


cyphar (approved):
LGTM, thanks!

PR #397, opened 2021-10-18
mutator: .Config() should return ispec.Image vs. ImageConfig

tych0:
We're interested in caching the full config (including diffids) outside of
umoci during mutation, but the mutator provides no access. Let's make it
so.

I'm open to other ways to do this, but this patch seems to be working fine for me.

### DIFF
--- a/cmd/umoci/config.go
+++ b/cmd/umoci/config.go
@@ -170,7 +170,7 @@ func config(ctx *cli.Context) error {
 		return errors.Wrap(err, "create mutator for manifest")
 	}
 
-	imageConfig, err := mutator.Config(context.Background())
+	config, err := mutator.Config(context.Background())
 	if err != nil {
 		return errors.Wrap(err, "get base config")
 	}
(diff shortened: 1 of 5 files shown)


cyphar:
Yeah I think the GHA runners no longer have `parallel` installed by default, so we need to update the CI to fix that...

cyphar:
It was necessary because without it you couldn't install `moreutils` because both packages provide a binary called `parallel`. `unlink` what StackOverflow told me to do to fix the issue. :shrug: 

If the GHA runners no longer have parallel installed, we can drop it entirely -- let's just hope they don't re-add it again in the future because it'll cause the builds to fail again.

cyphar:
The OSX CI issue should be fixed in #400.

tych0:
Ok, I rebased and dropped my fix. Hopefully it'll all work now...

cyphar:
I forgot to fix netlify to use `main` as well, lemme figure out how to re-trigger that...

cyphar:
I just force-pushed to your branch to kick it off again, hope you don't mind. Will merge once it's green.

tych0:
Cool, thanks, looks like it worked this time.

cyphar (approved):
LGTM.

PR #400, opened 2021-11-02
ci: osx: only try to unlink parallel if installed

cyphar:
It seems GHA changed their default installed set of brew packages to no
longer include parallel, which causes the "brew unlink" command to fail.
So first check if parallel is installed.


cyphar:
Ah the main branch name change has broken CI because "master" is referenced in the CI config. I will push the fix directly to master.

PR #402, opened 2021-11-10
Add CODEOWNERS

jdolitsky:


tych0:
I'm a little dubious about including this file; can we make it a symlink to maintainers or will that break automated tools?

Relevant discussion: https://groups.google.com/a/opencontainers.org/g/dev/c/w5SaT-p_m_k

cyphar:
Yeah, I'm not really sure this is necessary -- the MAINTAINERS file and teams are okay for now IMHO. If we want to switch all of OCI to CODEOWNERS then we can make that decision separately and umoci will follow along.

PR #404, opened 2021-11-19
oci: protect against mediatype confusion attacks (CVE-2021-41190)

cyphar:
Ensure that we never permit an ambigous blob (one that can be either a
manifest or index) and make sure we set and check the now-defined
MediaType field whenever parsing such blobs.

### DIFF
--- a/oci/casext/mediatype/parse.go
+++ b/oci/casext/mediatype/parse.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
 )
 
 // ParseFunc is a parser that is registered for a given mediatype and called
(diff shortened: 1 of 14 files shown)


oci/casext/mediatype/parse.go
```diff
@@ -136,22 +137,66 @@ func CustomJSONParser(v interface{}) ParseFunc {
...
+	if rdr != nil {
+		if err := json.NewDecoder(rdr).Decode(&index); err != nil {
+			return nil, err
+		}
+	}
+	if index.MediaType != "" && index.MediaType != ispec.MediaTypeImageIndex {
+		return nil, fmt.Errorf("malicious image detected: index contained incorrect mediaType: %s", index.MediaType)
+	}
+	if len(index.Config) != 0 {
+		return nil, fmt.Errorf("malicious image detected: index contained forbidden 'config' field")
```
tych0, line 166: For all these it seems like maybe we want errors.Errorf (and an errors.Wrapf()) so that users get stack traces?

oci/casext/mediatype/parse.go
```diff
@@ -136,22 +137,66 @@ func CustomJSONParser(v interface{}) ParseFunc {
 	if t.Kind() != reflect.Struct {
 		panic("CustomJSONParser() called with non-struct kind!")
 	}
-	return func(reader io.Reader) (_ interface{}, err error) {
+	return func(rdr io.Reader) (_ interface{}, err error) {
 		ptr := reflect.New(t)
-		if reader != nil {
-			err = json.NewDecoder(reader).Decode(ptr.Interface())
+		if rdr != nil {
+			err = json.NewDecoder(rdr).Decode(ptr.Interface())
```
tych0, line 143: we don't need this hunk, i guess?

cyphar: It was to make it consistent with the other methods (I've always preferred to use `rdr` anyway).

oci/casext/mediatype/parse.go
```diff
@@ -136,22 +137,66 @@ func CustomJSONParser(v interface{}) ParseFunc {
...
+func indexParser(rdr io.Reader) (interface{}, error) {
+	// Construct a fake struct which contains bad fields. CVE-2021-41190
+	var index struct {
+		ispec.Index
+		Config json.RawMessage `json:"config,omitempty"`
+		Layers json.RawMessage `json:"layers,omitempty"`
+	}
+	if rdr != nil {
+		if err := json.NewDecoder(rdr).Decode(&index); err != nil {
+			return nil, err
```
tych0, line 159: Still could use some errors.Wrapf()s around these, but I'll go ahead and land this since it's a CVE fix.

oci/casext/mediatype/parse.go
tych0: ok. might be worth a separate patch for it then.

tych0 (approved):

PR #407, opened 2021-12-09
pkg: turn down verified reader's complaints

tych0:
Simply traversing an image can yield:

  INFO[0001] verified reader: 116572160 bytes of trailing data discarded from vrdr[...]
  INFO[0002] verified reader: 116948992 bytes of trailing data discarded from vrdr[...]

Let's make this a debug so it generates less noise.

### DIFF
--- a/pkg/hardening/verified_reader.go
+++ b/pkg/hardening/verified_reader.go
@@ -188,7 +188,7 @@ func (v *VerifiedReadCloser) Close() error {
 		// If there's trailing bytes being discarded at this point, that
 		// indicates whatever you used to generate this blob is adding trailing
 		// gunk.
-		log.Infof("verified reader: %d bytes of trailing data discarded from %s", n, v.sourceName())
+		log.Debugf("verified reader: %d bytes of trailing data discarded from %s", n, v.sourceName())
 	}
 	// Piped to underlying close.
 	err := v.Reader.Close()


cyphar:
This should've been fixed by #375, so you shouldn't get anything when iterating through image layers anymore (we don't open a VerifiedReadCloser for media-types we know we can't parse during recursive walking).

The log message itself should not be silenced IMO -- it is meant to detect that the user has blobs which have garbage appended which is something they should know (this was added in response to an issue where someone had such an image which required us to read the rest of the blob).

tych0:
> media-types we know

I think that's the rub, is that we have unknown media types.

cyphar:
Sorry, I should've said "we only try to open a VerifiedReadCloser for media-types which we know we can parse". See the following chunk:

https://github.com/opencontainers/umoci/blob/5adbd995a49e596dfeae4178e3462a87566ec0c0/oci/casext/walk.go#L125-L131

cyphar:
Though that should be a `Debugf` log now that I think about it...

tych0:
> Can you give me a copy of the image so I can take a look at it?

Probably not, but if this behavior is not expected then I'll dig a bit deeper, maybe something else is going wrong.

cyphar:
If this is happening during a `casext.Walk`, I would say this is unexpected behaviour after #375.

tych0:
Ok, turns out this is just from a use of the FromDescriptor() API where we don't read the whole thing. I can close this for now, and we can remediate it some other way.

Issue #332, opened 2020-06-24
add cross-compiled binaries to releases


cyphar:
We need a `make cross` and we should cross-compile for releases.

thesayyn:
hey, @cyphar would you accept a PR for this? I believe it would be easier for me if I do this with goreleaser but I am not sure what your take would be on that.

cyphar:
For CI? That would probably be okay, though I suspect that it might be better to take the changes made to `release.sh` in github.com/opencontainers/runc -- I made quite a few updates which make my particular release workflow (build, sign locally using my smart card PGP key, upload).

thesayyn:
hmm, do you think it is worth creating this maintenance burden in this repository as well? I am happy to go that road but bash scripts are hard to maintain. 

goreleaser supports singing via pgp (see: https://goreleaser.com/customization/sign/). we could possibly make it the default pipeline for releases.

wdyt?

thesayyn:
@cyphar I have created a minimal patch for introducing goreleaser. It is almost identical to one that we have under `hack/release.sh`. though it missing `CGO_ENABLED=1` and i can’t figure out who provides `BUILDTAGS` env.

cyphar:
This was fixed in #518 and umoci 0.5.0 has cross-compiled binaries. I didn't notice that #505 was a duplicate of this.

PR #409, opened 2022-01-04
ci: cross compile and publish

thesayyn:
Addresses #332


cyphar:
You need to add a `Signed-off-by:` line to your commit(s) which indicates that you attest the [Developer Certificate of Origin](https://developercertificate.org/) a statement about your contributions that you *must read* before signing (don't worry, it's quite short and easy-to-read). You can add it to your commits with `git commit --amend -s`, and then doing a `git push --force`.

###### **NOTE**: This is a saved reply. Sorry if it reads as a cookie-cutter response, it was written so that newcomers understand what a "DCO" is and make the process for contributing a little less scary. ######

thesayyn:
Hey @cyphar, just a friendly ping. 

I have time to push this to get it landed. generated binaries seem to be working without any intervention. I have added platforms as I saw fit but I am afraid that we can’t support some of them due to the lack of a CI pipeline that tests for these platforms. Since you are the one who knows the codebase better could you take a look and point any errors that I am making? I would appreciate it a lot.

thesayyn:
@cyphar or anyone. (sorry for pinging again). could you PTAL? now, this is the pure bash script and uses some of the bits from runc repository.

thesayyn:
Remaining points: 
- Who is going to invoke scripts `build` and `sign`? Manually or by the CI.
- If latter, CI has to have the gpg private key present on the machine (if we are going to use [umoci.keyring](https://github.com/opencontainers/umoci/blob/main/umoci.keyring)) or one of its own.
- If again the latter, what's the trigger for release workflow? the [VERSION](https://github.com/opencontainers/umoci/blob/main/VERSION) file or a push to `refs/tag/x.x.x`?

cyphar:
Sorry for not taking a look earlier, I was on leave for a while and there were other issues I had to work on. Thanks for reworking this to be a shell script. I will take a closer look at how this interacts with my release workflow.

At the moment I do the build and signing on my local machine. I did think about coming up with a more complicated setup for both runc and umoci (trigger a binary build in CI which then creates a release with the unsigned artifacts, then run a locally script which pulls the artefacts, signs them, then re-uploads). But I'm not sure the time spent getting such a thing to work would be worth it.

As an aside, in runc, I recently reworked the signing system so that the binaries are all built in a container (so we can cross-compile them) and a separate script does the signing. For runc we needed to use a container because we need to build libseccomp from scratch and only Debian lets you have a full cross-compiling environment for all the platforms we need for runc. Since umoci is pure-Go I suspect these problems don't affect us as much.

thesayyn:
> EDIT: Ah, I see you switched to my new scripts from runc. That should make this simpler to review. Thanks!

yep. mostly comes from there. just changed it in a way that could cross-compile for different os and arch pairs. also, naming of the binaries changed drastically so there might be people who are depending on linux/amd64 binary that is already part of the releases. so it might be a good idea to have two copies of `linux/amd64` as `umoci_linux_amd64` and `umoci.amd64` so that we don't break any workflows out there downloading the binary with the old name.


> But I'm not sure the time spent getting such a thing to work would be worth it.
these scripts would work for your local release process. I don't see a strong reason you should change your release process either. 

Though, I believe it could be helpful to build unsigned (signed by CI) binaries for HEAD whenever HEAD moves forward. possibly as part of CI artifacts.  this could come in handy for testing unstable features and fixes out there in the wild. 
this could be a separate PR if you agree also.

> Since umoci is pure-Go I suspect these problems don't affect us as much.
Yeap. we don't need a container at all. compilation happens on the host machine.

PR #418, opened 2022-03-01
mutator: allow setting annotation data when generating a layer

tych0:
We want to include some additional hash data in the annotations when
generating layers, so let's allow setting this.

### DIFF
--- a/cmd/umoci/insert.go
+++ b/cmd/umoci/insert.go
@@ -190,7 +190,7 @@ func insert(ctx *cli.Context) error {
 
 	// TODO: We should add a flag to allow for a new layer to be made
 	//       non-distributable.
-	if _, err := mutator.Add(context.Background(), ispec.MediaTypeImageLayer, reader, history, mutate.GzipCompressor); err != nil {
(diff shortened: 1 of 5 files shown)


tych0:
Ping?

tych0:
Ping?

tych0:
Ping?

cyphar:
Ah, sorry -- I thought I'd already merged this. LGTM.

That being said, while having the ability to modify annotations is a good idea, we really should figure out some way of making annotations a better-handled aspect of umoci:

1. We should be able to modify annotations anywhere within the descriptor path (this is going to be non-trivial from a UX point of view and at least somewhat complicated from an API point of view). No other tool I'm aware of has the ability for you to do this, but we probably should try to do it so that others can make full use of annotations in an automated fashion.
2. Rewrite and expand `umoci stat` so you can properly inspect descriptor paths and their annotations (`umoci stat` was just a short hack I wrote for testing, we really need to have something better in its place).

Sadly whenever I've sat down to try to come up with a nice solution to the above, I couldn't really come up with one. It would also be nice to rework the mutator API a little bit so that we don't keep adding more and more function arguments.

cyphar (approved):
LGTM, though see my comment about what we might want to do in the future.

Issue #436, opened 2022-05-05
images created using umoci/layer can include layers that are invalid tar files


mikemccracken:
We have hit a situation where images built using stacker ( see https://github.com/project-stacker/stacker/blob/master/overlay/pack.go#L279 ) appear to be fine - they can be unpacked and run by runc, and umoci can extract them just fine - but the individual blobs are not all valid .tar.gz files.

here's the error seen in such a case:
```
$ zcat ../python3/blobs/sha256/91ae8c2f364ac6a5295646ad5b99fa95eaef3972691fca78482d6128e4c873eb | tar xf -
tar: Removing leading `/' from member names
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
```

the use case here is that security scanners that expect to be able to untar individual image layers are not able to do so, and our images get flagged as failing. 

Ideally we would never produce invalid tarballs, but failing that I'd like help understanding how this happens and how to be sure that the images aren't actually corrupt somehow.

I've attached a log from `umoci --log=debug unpack` of the image. here's the full log: [umoci-unpack-debug.log.gz](https://github.com/opencontainers/umoci/files/8636578/umoci-unpack-debug.log.gz)

the last few files that are unpacked are:
```
   • unpacking entry           path=var/log/apt/eipp.log.xz root=python3-1.0.9-2/rootfs type=48
   • unpacking entry           path=var/log/apt/history.log root=python3-1.0.9-2/rootfs type=48
   • unpacking entry           path=var/log/apt/term.log root=python3-1.0.9-2/rootfs type=48
   • unpacking entry           path=var/log/dpkg.log root=python3-1.0.9-2/rootfs type=48
   • ... done                 
```

if I untar the blob that contains these files, it stops with the EOF error above part-way through writing `var/log/dpkg.log` - it is clearly truncated - in this diff the `-` lines are from the tar-generated tree:
[umoci-unpack-debug.log.gz](https://github.com/opencontainers/umoci/files/8636578/umoci-unpack-debug.log.gz)

```
--- var/log/dpkg.log    2022-05-05 16:08:11.725576759 -0700
+++ ../python3-1.0.9/rootfs/var/log/dpkg.log    2022-04-28 15:31:59.000000000 -0700
@@ -97,4 +97,156 @@
 2022-04-28 22:30:45 trigproc libc-bin:amd64 2.31-0ubuntu9.7 <none>
[...]

PR #437, opened 2022-05-11
*: io.Copy improvements

cyphar:
* Implement EINTR-retry logic for `io.Copy` users to avoid erroring out for no reason.
* Improve io.Pipe CloseWithError usage so that we make any pipe-passed errors louder.

These changes might help with #436 but I'm not sure whether these were the root cause of the issue.

### DIFF
--- a/mutate/compress.go
+++ b/mutate/compress.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"runtime"
 
+	"github.com/apex/log"
 	zstd "github.com/klauspost/compress/zstd"
 	gzip "github.com/klauspost/pgzip"
 	"github.com/opencontainers/umoci/pkg/system"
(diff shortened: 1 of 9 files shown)


tych0:
I'm a little suspicious that we're hitting EINTR; most of these files we're writing to will have been created via the dir cas backend's PutBlob(), which it looks like uses ioutil.TempFile(), which doesn't open things as O_NONBLOCK AFAIK.

The patch itself looks fine to me, though.

cyphar:
@tych0 From a few Go releases ago, you can get `EINTR` randomly because Go now has far more aggressive preeimptive scheduling which results in loads of `EINTR`s popping up in IO operations. We had to switch to `EINTR` retry loops in runc and a few other projects. I don't know if that is the root cause (especially since they didn't see any errors during building) but I'd be surprised if this isn't something that might happen.

tych0:
Yeah, I've heard about this change. But man, that sucks. Really feels like you should not have to deal with EINTR from every possible function call that could cause a kernel sleep...

tych0 (approved):

cyphar:
Yeah I don't understand why they didn't handle this in the stdlib -- the behaviour of `io.Copy` should pretty clearly indicate that an EINTR should be ignored...

PR #447, opened 2022-07-07
test: handle /etc/pki/ca-trust with Fedora images

cyphar:
Fedora changed the file modes of /etc/pki/ca-trust to 000 which means
that during our unprivileged tests, trying to delete /etc doesn't work
(root can bypass the DAC rules but we can't even though we're the
owner).

The solution is to simply delete the files by recursively setting all
the directories as +rwx first, and then doing the recursive delete. In
order to avoid making this mistake with other tests in the future, add a
new helper (rm_rf) which does this.

### DIFF
--- a/test/gc.bats
+++ b/test/gc.bats
@@ -90,9 +90,8 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	bundle-verify "$BUNDLE"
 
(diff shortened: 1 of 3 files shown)


cyphar:
LGTM.

PR #467, opened 2023-04-04
chore: Update dependencies to latest for CVEs

blkperl:


blkperl:
@cyphar any tips on the test failures?

cyphar:
I'm working on the test failures. I started trying to fix them a few months ago but a bunch of other stuff got in the way. :sweat_smile:

cyphar:
There are two test failure issues. One is the lint one (working on that), the other is with this PR -- the updated vendor of the `https://github.com/rootless-containers/proto` contains a package with more than one Go `package ...` in it. I suspect there's something odd about what `go mod vendor` is doing...

cyphar:
Closing in favour of #470. Thanks for sending a PR though!

PR #499, opened 2023-09-13
Do not chmod +r if we don't need to

hallyn:
If the inode is a mountpoint, chmod may just fail.

(See https://github.com/project-stacker/stacker/issues/450)


cyphar:
I've just fixed the CI failure in main -- can you rebase?

hallyn:
Thanks, success :)

cyphar (approved):
LGTM

PR #501, opened 2023-09-17
ci: avoid go install for github.com/opencontainers/runtime-tool

cyphar:
All released versions of runtime-tool are not using go.mod, meaning that we end up pulling the latest version of their dependencies when using "go install". In particular, the runtime-spec update causes validation errors in our CI because we are still producing runtime-spec 1.0.2 config.jsons and runtime-tool doesn't support validating older versions.


cyphar:
LGTM, merging to fix CI.

PR #502, opened 2023-09-20
build(deps): bump github.com/opencontainers/runtime-spec from 1.0.3 t…

siretart:
…o 1.1.0-rc4

This version is currently used in Linux distributions to build podman and friends


cyphar:
There are other changes necessary to fully support image-spec 1.1, so while I appreciate the patch I will probably have to carry this myself (also, I want to do a release of umoci soon and releasing a version with an rc spec is something I'd like to avoid).

siretart:
By all means, please take over (or close) this PR! -- I needed that patch for Debian and am carrying this as a distro-patch for now

PR #506, opened 2023-10-04
tarGenerator.AddFile: ignore llistxattr eopnotsupp

hallyn:
If it's not supported, just assume there are no xattrs, don't fail the addfile.

### DIFF
--- a/oci/layer/tar_generate.go
+++ b/oci/layer/tar_generate.go
@@ -29,6 +29,7 @@ import (
 	"github.com/opencontainers/umoci/pkg/system"
 	"github.com/opencontainers/umoci/pkg/testutils"
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 )
 
 // ignoreXattrs is a list of xattr names that should be ignored when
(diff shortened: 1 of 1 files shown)


smoser (approved):
For some reference, we're hitting this when trying to create a new layer from a directory that contains a linux-kernel mounted squashfs image.

hallyn:
@cyphar  not sure why I had it in draft mode until monday - it is ready for review.

hallyn:
@tych0  ping

oci/layer/tar_generate.go
```diff
@@ -201,7 +202,10 @@ func (tg *tarGenerator) AddFile(name, path string) error {
 	// XXX: This should probably be moved to a function in tar_unix.go.
 	names, err := tg.fsEval.Llistxattr(path)
 	if err != nil {
-		return errors.Wrap(err, "get xattr list")
+		if err != unix.EOPNOTSUPP {
```
tych0, line 205: i think you want `errors.Cause()` here? something like this: https://github.com/opencontainers/umoci/blob/f74280bc7da2b5a7d0febd73b37278fe03fe6a17/oci/layer/tar_extract.go#L152-L154

hallyn: oh that's neat, thanks.

tych0 (approved):

cyphar:
FWIW, LGTM. :sweat_smile:

PR #508, opened 2023-10-11
fix(gzip): change compression block size

rchincha:
Several tools set this as the default value.

image/copy/copy.go
56:var compressionBufferSize = 1048576

skopeo/vendor/github.com/containers/image/v5/copy/compression.go
24:     compressionBufferSize = 1048576

Choosing a different value causes the final gzip'ed layer to become different although the source tar is identical, causing an content-addressable invariants to break.

### DIFF
--- a/mutate/compress.go
+++ b/mutate/compress.go
@@ -46,7 +46,7 @@ func (gz gzipCompressor) Compress(reader io.Reader) (io.ReadCloser, error) {
 	pipeReader, pipeWriter := io.Pipe()
 
 	gzw := gzip.NewWriter(pipeWriter)
-	if err := gzw.SetConcurrency(256<<10, 2*runtime.NumCPU()); err != nil {
+	if err := gzw.SetConcurrency(256<<12, 2*runtime.NumCPU()); err != nil {
 		return nil, errors.Wrapf(err, "set concurrency level to %v blocks", 2*runtime.NumCPU())
 	}
 	go func() {


hallyn (approved):

cyphar:
Unfortunately, a lot of things can cause the output blobs to change despite having the same input -- updates to the gzip compression library will cause changes, for instance (this is why we have tests for this in CI, and is the reason I haven't updated the gzip library we use -- but I think the data is too small to detect the difference in the buffer size).

> causing an content-addressable invariants to break

I'm not sure what invariants you're referring to -- there is no guarantee that you can get the same gzip blob from the same input with different programs/libraries (like all compression algorithms, gzip does not have a canonical representation). The best we've been able to do with umoci is try our best to guarantee that we will not produce different blobs for the same input across versions. (IMHO it seems pretty clear at this point that baking in compression to the image blob format was a mistake.)

Or are you running into this because you use umoci as a library as well as containers/image, and the different settings cause the gzip stream of the same input to have different outputs? That's annoying, but it seems possible for `github.com/klauspost/pgzip` to have non-deterministic output and so assuming that two gzip streams of the same input will have the same hash seems like a questionable assumption (presumably this is the issue you're having?).

(FWIW, I don't mind merging this, though it will mean that we will start outputting streams with different hashes to previous versions...)

rchincha:
> (IMHO it seems pretty clear at this point that baking in compression to the image blob format was a mistake.)
Without question -  anything that affects the final output should have been considered.

In the OCI world, unfortunately no single tool does all jobs and one has to be careful that they interoperate well. Our use case was to deduplicate content and unfortunately changing compression params changes the output although the input was identical.

As you may be already aware, we use umoci as a library in https://github.com/project-stacker/stacker.

> (FWIW, I don't mind merging this, though it will mean that we will start outputting streams with different hashes to previous versions...)

At least speaking just for us, we are ok with this ^

mikemccracken:
I think this might be better off as a configuration, there may indeed be cases where this will break assumptions in tools using this.

rchincha:
https://github.com/containers/image/commit/5b23c4b55a9e9d6fa59434a1b84fc9a734420610
^ Adding more context wrt to this issue.

rchincha:
@cyphar et al
Also pls look at: https://github.com/opencontainers/umoci/pull/509

rchincha:
Closing this in favor of PR #509

PR #509, opened 2023-10-13
fix(compress): allow passing in compressor options

rchincha:
Goals of this PR:
1. Allow passing in options to individual compressor
2. Do not change default behavior

### DIFF
--- a/mutate/compress.go
+++ b/mutate/compress.go
@@ -22,7 +22,14 @@ type Compressor interface {
 	// indicate what compression type is used, e.g. "gzip", or "" for no
 	// compression.
 	MediaTypeSuffix() string
+
+	// WithOpt applies an option and can be chained.
+	WithOpt(CompressorOpt) Compressor
 }
+
+// CompressorOpt is a compressor option which can be used to configure a
+// compressor.
+type CompressorOpt interface{}
 
 type noopCompressor struct{}
 
@@ -37,16 +44,24 @@ func (nc noopCompressor) MediaTypeSuffix() string {
 // NoopCompressor provides no compression.
 var NoopCompressor Compressor = noopCompressor{}
 
-// GzipCompressor provides gzip compression.
-var GzipCompressor Compressor = gzipCompressor{}
+func (nc noopCompressor) WithOpt(CompressorOpt) Compressor {
+	return nc
+}
 
-type gzipCompressor struct{}
+// GzipCompressor provides gzip compression.
+var GzipCompressor Compressor = gzipCompressor{blockSize: 256 << 10}
+
+type GzipBlockSize int
+
+type gzipCompressor struct {
+	blockSize int
+}
 
 func (gz gzipCompressor) Compress(reader io.Reader) (io.ReadCloser, error) {
 	pipeReader, pipeWriter := io.Pipe()
 
 	gzw := gzip.NewWriter(pipeWriter)
-	if err := gzw.SetConcurrency(256<<10, 2*runtime.NumCPU()); err != nil {
+	if err := gzw.SetConcurrency(gz.blockSize, 2*runtime.NumCPU()); err != nil {
 		return nil, errors.Wrapf(err, "set concurrency level to %v blocks", 2*runtime.NumCPU())
 	}
 	go func() {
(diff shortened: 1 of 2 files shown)


rchincha:
An alternative approach to: https://github.com/opencontainers/umoci/pull/508

rchincha:
https://github.com/project-stacker/stacker/blob/main/pkg/overlay/pack.go#L411C4-L411C4
becomes:

`GzipCompressor.WithOpt(GzipBlockLen(256<<12))`

hallyn:
Looks nice

rchincha:
https://github.com/project-stacker/stacker/pull/529

rchincha:
@cyphar thoughts on this PR? Hoping to avoid a umoci fork.

cyphar:
A fork definitely won't be necessary, I'll review this on Monday. Sorry, I was a bit busy last week.

mutate/compress.go
```diff
@@ -76,6 +91,15 @@ func (gz gzipCompressor) MediaTypeSuffix() string {
 	return "gzip"
 }
 
+func (gz gzipCompressor) WithOpt(opt CompressorOpt) Compressor {
+	switch val := opt.(type) {
```
cyphar, line 95: I'm not sure I like that we ignore invalid arguments here. If we make this act on `*gzipCompressor` we could make it return a bool (or error) and then the caller can decide if they care about whether `WithOpt` actually did anything.

Alternatively, it feels like it should be possible to reimplement all of this compressor stuff with a closure that is passed by the caller (that creates the writer and configures it). But that would require more work...

cyphar: Maybe we could add a `CustomCompressor` just for cases where you need to configure the compression? Idk...

rchincha:
Another place of relevance in umoci code wrt compressor options :(

https://github.com/opencontainers/umoci/blob/main/pkg/system/copy.go#L32

cyphar:
Given #562 this would need to be reworked, but I don't know if you still have interest in this PR @rchincha? I can carry it if you like. I suspect we would want to make it more like the `(... , opts ..func(...))`-style configuration pattern now that algorithms are properly supported.

mikemccracken:
> @cyphar stacker is using this, yes, and AFAIK it is because a user specifically need it.

stacker has this commit in its stale fork of umoci, yes.

> @mikemccracken can you confirm whether the ability to specify the compression options are still needed?

For background, here is my summary of how this became necessary, lightly edited from an internal note in 2023:

	•	docker and umoci  have different gzip buffer sizes, and **docker build re-compresses base image layers** (this part is weird, but maybe it's because of supporting docker commit).
	•	we are seeing this (layers that should be identical not matching) because some app teams are doing docker builds on top of stacker-created base image layers so the base layers don't match after docker recompresses them
	•	the Right Fix is to have everyone use stacker, but that won't happen this cycle and we need a fix for this cycle. (note, this is from many cycles ago and nothing has changed)
	•	changing stacker to use the same buffer size as docker should be fine and won't cause too much coordination between users of stacker because stacker does not recompress base layers, so old stacker versions can still build with new-stacker created base layers and it won't cause problems

other tools like the containers/image library also match the docker compression option, so they don't have this problem.

----

I believe that the people who were using `docker build` on top of stacker-built images are still doing so, but I don't know for sure. I do think that it'd come up again if we stopped this but I'm also a little surprised it hasn't come up in other uses of umoci.

Q: Would it make sense to just hardcode in umoci the same compression block size as docker and containers/image use?
It's a workaround for a weird docker behavior but it also seems basically harmless, IIRC the difference was not big vs. the default.

mikemccracken:
> So, ideally we wouldn't change the current default because doing so would cause the same unfortunate cache busting. However, if Docker really does break our layers then maybe we should just hard-code it...

I don't follow what you mean by cache busting here - AFAIK the only reason this parameter ever comes up is that docker recompresses base tarballs using its own hardcoded blocksize and expects to get the same bits. Are there other tools also doing this?

> (It is possible to make the new algorithm setup configurable but it's probably not worth the effort unless there's a strong need for it.)

As for configurability, stacker only needs to match docker, if umoci's default is the same as stacker then we don't need configurability here.

cyphar:
Unless I'm missing something, this seems to actually be an implementation detail of how containerd buffers their writes to the OCI store (`bufPool` in `github.com/containerd/containerd/content/local/store.go`).

cyphar:
#581 changes our default to the one used by stacker. PTAL @rchincha.

PR #510, opened 2023-10-17
mutate: add uncompressed blob size annotation

mikemccracken:
Sometimes it is useful to have a reliable estimate of the space that would be taken up by a layer when it is unpacked. Since gzip's headers are only accurate up to 4GiB, larger layers can not be easily measured without uncompressing them.

This commit adds a field to the gzip and zstd compressors to store the bytes read from the stream during compression and uses that to set an annotation on newly generated layers.

If the noop compressor is used, the annotation is not added, as the existing "compressed" layer size would be the same.

👀 note that the annotation name is just a shot in the dark, I am open to suggestions for something that'd make more sense.

For validation, I unpacked the `opensuse/leap:15.4` image with `sudo ~/umoci/umoci unpack --image opensuse:latest opensuse-latest`

added a 5GiB file with just `truncate -s 5GB opensuse-latest/rootfs/test`

 then repacked and inspected the resulting blob:
`sudo ~/umoci/umoci --log=debug repack --image ./opensuse:latest-repacked-w-bigchange opensuse-latest/`

here are the layers from the resulting image:
```json
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:97fb1fedf685704ab0caadb3872223dfa7a49e26e65ea3a57aa34508d93ea5b8",
      "size": 44987689
    },
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:cb6f4fa41f115bc6070e91509cc23a5df6a98da11c971060eb33a74a94861362",
      "size": 27975432,
      "annotations": {
        "ci.umo.uncompressed_blob_size": "5368711168"
      }
    }
  ]
  ```
  
  and we can see the incorrect gzip header, followed by the actual size after decompressing:
  (note I copied the blobs from the layout into this tmp dir, this is slightly edited)

```
embrane@atom-lab-8:/tmp/umoci$ file ../opensuse/blobs/sha256/cb6f4fa41f115bc6070e91509cc23a5df6a98da11c971060eb33a74a94861362
../opensuse/blobs/sha256/cb6f4fa41f115bc6070e91509cc23a5df6a98da11c971060eb33a74a94861362: gzip compressed data, original size modulo 2^32 1073743872
# elide copy and renaming blobs here
embrane@atom-lab-8:/tmp/umoci$ gunzip cb6f4fa41f115bc6070e91509cc23a5df6a98da11c971060eb33a74a94861362.tar.gz
embrane@atom-lab-8:/tmp/umoci$ ll
total 5242888
drwxrwxr-x   3 embrane embrane        100 Oct 16 16:52 ./
drwxrwxrwt 328 root    root         11560 Oct 16 16:50 ../
-rw-------   1 embrane embrane 5368711168 Oct 16 16:51 cb6f4fa41f115bc6070e91509cc23a5df6a98da11c971060eb33a74a94861362.tar
```

### DIFF
--- a/mutate/compress.go
+++ b/mutate/compress.go
@@ -22,6 +22,10 @@ type Compressor interface {
 	// indicate what compression type is used, e.g. "gzip", or "" for no
 	// compression.
 	MediaTypeSuffix() string
+
+	// BytesRead returns the number of bytes read from the uncompressed input
+	// stream at the current time, no guarantee of completion.
+	BytesRead() int64
 }
 
 type noopCompressor struct{}
@@ -34,15 +38,21 @@ func (nc noopCompressor) MediaTypeSuffix() string {
 	return ""
 }
 
+func (nc noopCompressor) BytesRead() int64 {
+	return -1
+}
+
 // NoopCompressor provides no compression.
 var NoopCompressor Compressor = noopCompressor{}
 
 // GzipCompressor provides gzip compression.
-var GzipCompressor Compressor = gzipCompressor{}
+var GzipCompressor Compressor = &gzipCompressor{}
 
-type gzipCompressor struct{}
+type gzipCompressor struct {
+	bytesRead int64
+}
 
-func (gz gzipCompressor) Compress(reader io.Reader) (io.ReadCloser, error) {
+func (gz *gzipCompressor) Compress(reader io.Reader) (io.ReadCloser, error) {
 	pipeReader, pipeWriter := io.Pipe()
 
 	gzw := gzip.NewWriter(pipeWriter)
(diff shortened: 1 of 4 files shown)


mikemccracken:
now that CI is passing here, ping @cyphar or @tych0 , please take a look, and especially let me know if `ci.umo.uncompressed_blob_size` is the right key to use for this. Thanks!

tych0 (approved):
lgtm... also seems very useful. why did the original spec not have this?

rchincha:
> lgtm... also seems very useful. why did the original spec not have this?

A very very good question which @cyphar also pointed to and we are painfully finding out when inter-operating with other tools :(

hallyn:
Uh, hm.  Ping @tych0 ?  Was this meant to be merged?

tych0:
I don't see why not, thanks for the ping.

PR #527, opened 2024-01-23
fix a Sirupsen/logrus import issue in CI

tych0:
we are seeing:

2.947 go: github.com/opencontainers/image-tools/cmd/oci-image-tool imports
2.947 	github.com/Sirupsen/logrus: github.com/Sirupsen/logrus@v1.9.3: parsing go.mod:
2.947 	module declares its path as: github.com/sirupsen/logrus
2.947 	        but was required as: github.com/Sirupsen/logrus

in CI, hopefully this replace fixes it?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN zypper -n in \
 		curl \
 		git \
 		gnu_parallel \
-		"go>=1.18" \
+		"go==1.21" \
 		go-mtree \
 		gzip \
 		jq \


tych0:
that wasn't it, trying something else...

tych0:
@cyphar i'm not sure if zypper has a "don't include prereleases" option, that's probably better than just hard coding 1.21, since we'll have to keep bumping it forever...

cyphar:
The reason is that our packaging in openSUSE has (had?) a bug where trying to install just "go" didn't give you the latest version. It should've been fixed, but since I handed off the Go packaging to a different guy at SUSE a few years ago it's been harder to get answers to these kinds of questions (I used to do Go packaging for (open)SUSE). We can try to remove and see what happens...

go.mod
```diff
@@ -39,3 +39,7 @@ require (
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace (
+    github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.9.0
```
cyphar, line 44: I'm confused why this has come up all of a sudden -- the Sirupsen -> sirupsen change happened many years ago and we don't have any `Sirupsen/logrus` import paths in the codebase...

If we do have any of the old import paths, can we just rewrite them?

Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN zypper -n in \
 		curl \
 		git \
 		gnu_parallel \
-		"go>=1.18" \
+		"go==1.21" \
```
cyphar, line 35: You've changed from `>=` to `==`. :sweat_smile:

go.mod
tych0: sorry, this was just noise from my previous attempt. the golang thing is all that's necessary.

Dockerfile
cyphar: Oh, that's the issue? That is _very_ weird. I'll need to look into it...

cyphar (approved):
LGTM, just to fix CI. It is very concerning that Go 1.22 might break working projects?

Also I'm confused why there is a pre-release version of Go in the repos... I need to speak to the Go maintainer for openSUSE...

tych0:
yeah, i'm hopeful it's just a regression.

PR #534, opened 2024-02-22
generateinsertlayer: ensure that we close the tarwriter

### LINKED ISSUE #436: images created using umoci/layer can include layers that are invalid tar files

mikemccracken:
This fixes #436.

I added a test that shows the issue - the commit with the fix has the full explanation, repeated here:

generatelayer closes the tarwriter, but generateinsertlayer forgets to.

Closing the tarwriter writes the required footer of 1k of zeros.

This results in tar files that are complete but invalid, and different
reading tools will behave differently:

- bsdtar doesn't complain and exits 0
- gnu tar (and security scanning tools that use it) will exit 2 with an
unexpected EOF message
- python's tarfile library will raise an Unexpected EOF error
- golang's archive/tar library can raise an unexpected EOF error, but
  for some files created by generateinsertlayer, it just raises a
  plain EOF error, which means golang based tools generally ignore this
  and work fine, this includes umoci.

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -99,7 +99,21 @@ function requires() {
 }
 
 function image-verify() {
-	oci-image-tool validate --type "imageLayout" "$@"
+	local ocidir="$@"
+	# test that each generated targz file is valid according to gnutar:
+	for f in $(ls $ocidir/blobs/sha256/); do
+		file $ocidir/blobs/sha256/$f | grep "gzip" || {
+			continue
+		}
+		zcat $ocidir/blobs/sha256/$f | tar tvf - >/dev/null || {
+			rc=$?
+			file $ocidir/blobs/sha256/$f
+			echo "error untarring $f: $rc"
+			return $rc
+		}
+		echo $f: valid tar archive
+	done
+	oci-image-tool validate --type "imageLayout" "$ocidir"
 	return $?
 }
 
--- a/oci/layer/generate.go
+++ b/oci/layer/generate.go
@@ -54,7 +54,9 @@ func GenerateLayer(path string, deltas []mtree.InodeDelta, opt *RepackOptions) (
 	go func() (Err error) {
 		// Close with the returned error.
 		defer func() {
-			log.Warnf("could not generate layer: %v", Err)
+			if Err != nil {
+				log.Warnf("could not generate layer: %v", Err)
+			}
 			// #nosec G104
 			_ = writer.CloseWithError(errors.Wrap(Err, "generate layer"))
 		}()
@@ -135,7 +137,9 @@ func GenerateInsertLayer(root string, target string, opaque bool, opt *RepackOpt
 
 	go func() (Err error) {
 		defer func() {
-			log.Warnf("could not generate insert layer: %v", Err)
+			if Err != nil {
+				log.Warnf("could not generate insert layer: %v", Err)
+			}
 			// #nosec G104
 			_ = writer.CloseWithError(errors.Wrap(Err, "generate insert layer"))
 		}()
--- a/oci/layer/generate.go
+++ b/oci/layer/generate.go
@@ -142,6 +142,12 @@ func GenerateInsertLayer(root string, target string, opaque bool, opt *RepackOpt
 
 		tg := newTarGenerator(writer, packOptions.MapOptions)
 
+		defer func() {
+			if err := tg.tw.Close(); err != nil {
+				log.Warnf("generate insert layer: could not close tar.Writer: %s", err)
+			}
+		}()
+
 		if opaque {
 			if err := tg.AddOpaqueWhiteout(target); err != nil {
 				return err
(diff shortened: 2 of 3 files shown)


tych0:
Everything looks good to me, except you have,

> (cherry picked from commit 3f53b10037f03d01412b5b0eb9e50a0053b7a156)

in your commit message, which is a mystery hash?

tych0:
I just cleaned it up, will merge once that all passes CI

tych0 (approved):

test/helpers.bash
```diff
@@ -99,7 +99,21 @@ function requires() {
 }
 
 function image-verify() {
-	oci-image-tool validate --type "imageLayout" "$@"
+	local ocidir="$@"
+	# test that each generated targz file is valid according to gnutar:
+	for f in $(ls $ocidir/blobs/sha256/); do
+		file $ocidir/blobs/sha256/$f | grep "gzip" || {
```
cyphar, line 105: Nit: It would be nice if this check was more "strictly correct" (i.e. it actually detected if the file was actually a tar archive rather than a gzip stream) but given this is just for tests that's not too important.

cyphar (approved):
LGTM, good catch!

mikemccracken:
> Everything looks good to me, except you have,
> 
> > (cherry picked from commit 3f53b10037f03d01412b5b0eb9e50a0053b7a156)
> 
> in your commit message, which is a mystery hash?

d'oh- thanks for taking care of this, that hash was a commit on top of the stacker fork of umoci, I should've cleaned up the message.

Thanks for the quick review guys!

test/helpers.bash
mikemccracken: I agree... I actually thought maybe the right thing was to have the oci-image-tool do this validation, but that seems to be abandoned.
Is there a different tool for standards validation out there somewhere that we should move this to?
Given the recent changes with artifacts, etc, it is probably extra stale now.

cyphar: There were discussions about moving the image-validation infrastructure to umoci, but nothing has happened with regards to that topic since then. I can take a look at doing this after I do an umoci 0.5 release.

PR #538, opened 2024-03-29
fix: handle overlay xattr opaque bit

rchincha:
Current behavior determines if a path is a whiteout if a overlay char dev is present.

Additionally, also check the extended attrs.

### DIFF
--- a/oci/layer/utils.go
+++ b/oci/layer/utils.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/apex/log"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/umoci/pkg/fseval"
 	"github.com/opencontainers/umoci/pkg/idtools"
 	"github.com/pkg/errors"
 	rootlesscontainers "github.com/rootless-containers/proto/go-proto"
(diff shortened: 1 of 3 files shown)


hallyn (approved):

hallyn:
Looks like some delta between current main branch and the project-stacker branch, @rchincha .

hallyn (approved):

tych0:
This behavior seems a bit fiddly; can you add a test to make sure it doesn't break in the future? I think that will also help describe what the problem is.

cyphar:
Sorry for not responding earlier when you sent this -- my first impression when I looked at this a few months ago is that we might want to look at what containerd does, since they have to do conversions between the AUFS-style OCI tar entries to overlayfs files. But I need to look at this again, since I might be confusing what the issue is...

oci/layer/utils.go
```diff
@@ -243,6 +244,25 @@ func isOverlayWhiteout(info os.FileInfo) (bool, error) {
...
 
-	return major == 0 && minor == 0 &&
-		info.Mode()&os.ModeCharDevice != 0, nil
+	if major == 0 && minor == 0 &&
+		info.Mode()&os.ModeCharDevice != 0 {
+		return true, nil
+	}
+
+	// also evaluate xattrs which may have opaque value set
+	attr, err := fsEval.Lgetxattr(fullPath, "user.overlay.opaque")
```
cyphar, line 253: Seems to me like this should be `trusted.overlay.opaque`.

oci/layer/utils.go
```diff
@@ -243,6 +244,25 @@ func isOverlayWhiteout(info os.FileInfo) (bool, error) {
...
+	if err != nil {
+		v := errors.Cause(err)
+		if !errors.Is(err, os.ErrNotExist) && v != unix.EOPNOTSUPP && v != unix.ENODATA && v != unix.EPERM && v != unix.EACCES {
+			return false, errors.Errorf("[internal error] unknown stat info type %T", info.Sys())
+		}
```
cyphar, line 258: This error handling doesn't make much sense to me, I guess you copied it from above?

cyphar:
#572 implements this properly as well as fixing a few other aspects of our overlayfs handling.

PR #556, opened 2024-11-28
*: migrate to Go stdlib error wrapping

### LINKED ISSUE #388: Switch to Go 1.13 error types
So we don't need to use `github.com/pkg/errors` anymore. Unfortunately this'll mean we can't get stack traces anymore but I don't think that'll be the end of the world.

cyphar:
Fixes #388


tych0 (approved):

cyphar:
Yeah, it turns out depending on the behaviour of `errors.Wrap(nil, ...)` was a bad idea...

cyphar:
This patchset adds and removes `pkg/fmtcompat`. I think that it makes it easier to review, so I think it makes sense to keep it (otherwise you would need to do everything in one big patch or rework things in a somewhat unnatural way, otherwise you'll end up with broken commits if you bisect).

PR #562, opened 2025-04-03
more generic and full zstd compression support

cyphar:
Make our compression handling more generic, and support zstd decompression.

Also add a `--compress` flag that lets users specify compression for new layers. The default configuration is `--compress=auto` which (in contrast to the previous hardcode-gzip behaviour) will try to use the same compression algorithm as in the last layer. See `umoci-repack(1)` for more information.

### DIFF
--- a/pkg/iohelpers/count_reader.go
+++ b/pkg/iohelpers/count_reader.go
@@ -1,0 +1,48 @@
+/*
+ * umoci: Umoci Modifies Open Containers' Images
+ * Copyright (C) 2016-2025 SUSE LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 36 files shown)


pkg/iohelpers/count_reader.go
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package iohelpers
+
+import (
+	"io"
+)
+
+// An io.Reader wrapper that counts how many bytes were read from the
```
tych0, line 24: `read from the` what? :)

tych0 (approved):
Overall looks pretty reasonable to me.

pkg/iohelpers/count_reader.go
cyphar: Oops, fixed. It seems that our `golint` is not running properly either, since there were some other `golint` issues I caught when running locally...

PR #572, opened 2025-04-25
layer: numerous overlayfs whiteout fixes

cyphar:
It turns out that a lot of our overlayfs whiteout conversion code is actually quite broken, and quite a few things are not properly supported. Here are the set of issues we needed to fix:

- [x] `GenerateLayer` would produce whiteouts with the wrong pathname (it would use the full path rather than the layer path), leaking information about the host and producing incorrect images as output.
- [x] `GenerateLayer` would skip any non-whiteouts files when operating in `TranslateOverlayfsWhiteout` mode, silently losing data.
- [x] Our testing of `TranslateOverlayfsWhiteout` was quite crude.
- [x] The `GenerateInsertLayer` code would not use `fsEval` in most places, which would cause rootless mode to not work properly.
- [x] We need to support converting `trusted.overlay.opaque=y` since users might use that to specify opaque whiteouts (there is no other way to specify this when umoci generates archives).
  - This is a reimplementation of #538 (which had several issues).
- [x] Extracting overlayfs whiteouts doesn't work if there are missing parent directories or the filetype of the opaque dir is wrong. In practice this is not an issue with most images but for completeness we should handle this case properly.
- [x] Currently extracting regular whiteouts in an opaque whiteout directory will lead to confusing effects with `overlayfs` (`readdir` will report the existence of files that are whited out).
  - We will probably need to switch `(*tarExtractor).upperPaths` to be a trie that will allow us to remove already-extracted regular whiteouts from a directory that then gets tagged as an opaque whiteout, as well as skip extracting regular whiteouts from a directory that is already and opaque whiteout. None of this is necessary for regular extractions but overlayfs kind of forces us to do this.
- :twisted_rightwards_arrows: `user.overlay.*` support feature moved to #576.
- [x] We need to support the `trusted.overlay.` escaping so that:
  - Repacking a regular non-overlayfs layer with `trusted.overlay.*` will keep the same xattr name.
  - Repacking an overlayfs layer with `trusted.overlay.*` will strip the xattrs but `trusted.overlay.overlay.*` will have one layer of escaping stripped when generating the tar archive.
  - Ditto when unpacking -- for non-overlayfs we extract as normal but for overlayfs we need to add escaping.

### DIFF
--- a/oci/layer/generate.go
+++ b/oci/layer/generate.go
@@ -68,7 +68,7 @@ func GenerateLayer(path string, deltas []mtree.InodeDelta, opt *RepackOptions) (
 		// We can't just dump all of the file contents into a tar file. We need
 		// to emulate a proper tar generator. Luckily there aren't that many
 		// things to emulate (and we can do them all in tar.go).
-		tg := newTarGenerator(writer, packOptions.MapOptions)
+		tg := newTarGenerator(writer, packOptions)
 
 		// Sort the delta paths.
 		// FIXME: We need to add whiteouts first, otherwise we might end up
@@ -150,7 +150,7 @@ func GenerateInsertLayer(root string, target string, opaque bool, opt *RepackOpt
 			_ = writer.CloseWithError(closeErr)
 		}()
 
-		tg := newTarGenerator(writer, packOptions.MapOptions)
+		tg := newTarGenerator(writer, packOptions)
 
 		defer func() {
 			if err := tg.tw.Close(); err != nil {
--- a/oci/layer/generate.go
+++ b/oci/layer/generate.go
@@ -96,7 +96,8 @@ func GenerateLayer(path string, deltas []mtree.InodeDelta, opt *RepackOptions) (
 						return err
 					}
 					if whiteout {
-						if err := tg.AddWhiteout(fullPath); err != nil {
+						log.Debugf("generate layer: converting overlayfs whiteout %q to OCI whiteout", name)
+						if err := tg.AddWhiteout(name); err != nil {
 							return fmt.Errorf("generate whiteout from overlayfs: %w", err)
 						}
 					}
@@ -181,7 +182,7 @@ func GenerateInsertLayer(root string, target string, opaque bool, opt *RepackOpt
 				return err
 			}
 			if packOptions.TranslateOverlayWhiteouts && whiteout {
-				log.Debugf("converting overlayfs whiteout %s to OCI whiteout", name)
+				log.Debugf("generate insert layer: converting overlayfs whiteout %q to OCI whiteout", name)
 				return tg.AddWhiteout(name)
 			}
 
--- a/oci/layer/generate.go
+++ b/oci/layer/generate.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 	"path/filepath"
 	"sort"
 
@@ -166,22 +165,27 @@ func GenerateInsertLayer(root string, target string, opaque bool, opt *RepackOpt
 		if root == "" {
 			return tg.AddWhiteout(target)
 		}
-		return unpriv.Walk(root, func(curPath string, info os.FileInfo, err error) error {
+		return unpriv.Walk(root, func(fullPath string, info os.FileInfo, err error) error {
 			if err != nil {
 				return err
 			}
 
-			pathInTar := path.Join(target, curPath[len(root):])
+			relName, err := filepath.Rel(root, fullPath)
+			if err != nil {
+				return err
+			}
+			name := filepath.Join(target, relName)
+
 			whiteout, err := isOverlayWhiteout(info)
 			if err != nil {
 				return err
 			}
 			if packOptions.TranslateOverlayWhiteouts && whiteout {
-				log.Debugf("converting overlayfs whiteout %s to OCI whiteout", pathInTar)
-				return tg.AddWhiteout(pathInTar)
+				log.Debugf("converting overlayfs whiteout %s to OCI whiteout", name)
+				return tg.AddWhiteout(name)
 			}
 
-			return tg.AddFile(pathInTar, curPath)
+			return tg.AddFile(name, fullPath)
 		})
 	}()
 	return reader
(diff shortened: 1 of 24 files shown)


oci/layer/generate.go
```diff
@@ -87,21 +90,36 @@ func GenerateLayer(path string, deltas []mtree.InodeDelta, opt *RepackOptions) (
...
+								log.Warnf("generate layer: could not add directory entry for opaque from overlayfs for file %q: %s", name, err)
+								return fmt.Errorf("generate directory entry for opaque whiteout from overlayfs: %w", err)
+							}
+							err = tg.AddOpaqueWhiteout(name)
+						default:
+							return fmt.Errorf("[internal error] unknown overlayfs whiteout type %q", woType)
+						}
+						if err != nil {
+							log.Warnf("generate layer: could not add whiteout %s from overlayfs for file %q: %s", woType, name, err)
+							return fmt.Errorf("generate whiteout %s from overlayfs: %w", woType, err)
```
tych0, line 119: Is there a reason to both warn and error here and above?

oci/layer/utils_unix.go
```diff
@@ -0,0 +1,94 @@
...
+
+type overlayWhiteoutType string
+
+const (
+	overlayWhiteoutPlain  overlayWhiteoutType = "plain whiteout"
+	overlayWhiteoutOpaque overlayWhiteoutType = "opaque whiteout"
+)
+
+// isOverlayWhiteout returns true if the FileInfo represents an overlayfs style
+// whiteout (i.e. mknod c 0 0) and false otherwise.
```
tych0, line 40: I think this comment is pretty wrong now :)

oci/layer/utils_unix.go
```diff
@@ -0,0 +1,94 @@
...
+	case unix.S_IFCHR:
+		// classic char 0:0 style whiteouts
+		if stat.Rdev == 0 {
+			return overlayWhiteoutPlain, true, nil
+		}
+	case unix.S_IFDIR:
+		// opaque whiteouts
+		val, err := fsEval.Lgetxattr(path, "trusted.overlay.opaque")
+		if err != nil {
+			// If we are missing privileges to read the xattr (ENODATA) or the
```
tych0, line 57: Is ignoring ENODATA safe? If someone has marked an `/etc/foo.d` directory as a whiteout, lower layers might be able to inject config that higher layers tried to delete. Or am I misunderstanding?

oci/layer/xattr.go
```diff
@@ -0,0 +1,214 @@
...
+		return &xattr
+	}
+	escaped := "trusted.overlay.overlay." + subXattr
+	return &escaped
+}
+
+func (_ overlayXattrFilter) ToTar(woMode WhiteoutMode, xattr string) *string {
+	// If we are not in overlayfs extraction mode, we can just pack the xattr
+	// as normal. However, this is a bit of a strange situation so we should
+	// output a warning.
```
tych0, line 146: Hm? I guess it's not that strange, isn't it the default?

tych0 (approved):

oci/layer/xattr.go
cyphar: This comment probably should be removed now that we no longer log anything here, but this is kind of unusual -- you wouldn't expect to see `trusted.overlay.foo` xattrs in a regular image or a regular rootfs. I would assume it's a configuration error honestly (you forgot to specify the on-disk format is overlayfs).

oci/layer/utils_unix.go
cyphar: My understanding of the usecase is that you wouldn't give untrusted user access to the raw overlayfs-extracted lowerdir, you would give them access to a merged overlayfs right? You can't set those xattrs through a merged overlayfs AFAIK.

(Also for the moment it's `trusted.overlay.opaque` so an unprivileged user couldn't write to it anyway. But this is the concern I had with enabling `user.overlay.*` support by default without making it a config option.)

oci/layer/generate.go
cyphar: We do this in a couple of places, the main issue is that I have seen cases where `CloseWithError` doesn't end up propagating the error properly and so this is meant to be a "final fallback" to tell CLI users that something was up rather than succeeding silently.

I don't know if the `CloseWithError` thing has been fixed. It was a weird race condition that I couldn't ever really reproduce.

oci/layer/xattr.go
tych0: What if the filesystem has overlay whiteouts applied though? Like in stacker's case, if someone does a delete which generates an overlay whiteout xattr, but you're converting back to tar, won't this code get called?

But in any case, I agree that just deleting the comment is fine.

oci/layer/utils_unix.go
tych0: > My understanding of the usecase is that you wouldn't give untrusted user access to the raw overlayfs-extracted lowerdir, you would give them access to a merged overlayfs right?

Maybe, but in stacker's case it will extract them with the unprivileged uid/gid, so it's not impossible. Though I agree, it requires a container escape, and at that point maybe it doesn't matter.

oci/layer/xattr.go
cyphar: No, in that case it gets handled in `oci/layer/generate.go` (where `AddWhiteout` gets called instead of `Add`), same as before.

oci/layer/utils_unix.go
cyphar: I think even in that case you'd be mounting the overlayfs with `-o userxattr` and whatever untrusted program running atop the rootfs won't be able to set the xattrs either.

For an attacker to access the lowerdirs I think you'd need a container escape, as you said.

cyphar: FWIW, I would prefer it if we could avoid this by just checking `llistxattr` (like with `whiteout`) but sadly that's not how that xattr works...

But I also think that `ENODATA` here is ultimately fine because it just means that either umoci isn't privileged enough (we need `userxattr` support to fix this -- #576) or the rootfs has an empty xattr (which could happen for a variety of reasons, and an attacker that can do that can do plenty of other bad things imho).

hallyn (approved):
(some offline in-depth discussion later...)

lgtm, thanks.

I'm hoping to test stacker against this later this week.  Will open issues if I run into anything, but if there are issues i don't believe they'd be regressions, rather things this didn't fix (which the stacker fork did).

PR #580, opened 2025-05-07
oci: layer: temporarily disallow overlayfs on-disk bundle unpacking

cyphar:
We currently do not properly support this, and the results users get
will be completely broken. We can re-enable this in the future.

The only user of OverlayFSWhiteout is stacker, and they use UnpackLayer
explicitly (without using our bundle unpacking logic at all).


cyphar:
@tych0 @hallyn This is the breaking changes patch I mentioned. Stacker does not use `UnpackRootfs` nor `UnpackManifest` (and the current implementation is broken enough that you would've complained about it by now), but I wanted to split this out so it could be reviewed by itself and not get lost in #572.

tych0 (approved):

PR #587, opened 2025-05-20
oci: layer: add support for userxattr in OverlayfsRootfs

### LINKED ISSUE #576: layer: overlay userxattr support
Moved out of #572 because it ultimately may require some of the `umoci.json` changes mentioned in https://github.com/opencontainers/umoci/issues/574#issuecomment-2832639560 (and #584) and is not a critical bugfix for 0.5.

<hr>

overlayfs added support for unprivileged mounting a long time ago, and the `userxattr` option allows unprivileged users to create whiteouts (`mknod c 0 0` works for unprivileged users for a while now, but `trusted.overlay.opaque` doesn't) and configure other overlayfs xattrs.

However, mounting overlayfs without `userxattr` will result in said `user.overlay.*` xattrs becoming visible, which is not something we want. This needs to be configurable so that the overlayfs mount is aware that we do or don't have `userxattr`-compatible xattrs (probably as a broader part of #574).

> - [ ] We need to find a nice way of supporting the `user.overlay.*` namespace.
>   - For `Generate*Layer` we should arguably generate overlayfs regardless of which namespace is used but when extracting we should probably have a config option to specify what namespace we should use (or both). The default could be based on whether we are rootless or not.

### LINKED ISSUE #584: layer pack/unpack: unify on-disk format configuration
Currently we have two different systems for configuring the on-disk format for packing (`TranslateOverlayWhiteouts`) and unpacking (`WhiteoutMode`), but ideally it would be a single configuration type with per-format configuration (in the case of overlayfs this would be whether to use userxattr -- see #576).

cyphar:
Since Linux 5.11, overlayfs supports using "user.overlay." as the xattr
namespace for special overlayfs xattrs (rather than the default
"trusted.overlay.") when mounted using the "userxattr" mount option.
This is a key part of unprivileged overlayfs mounting.

For the most part, supporting this is fairly straightforward now that
the on-disk format is represented with OverlayfsRootfs -- all hardcoded
references to "trusted.overlay." just need to take into account the
value of OverlayfsRootfs.UserXattr. Downstreams like stacker have had
out-of-tree patches to support operating on these xattrs for a long time
(since they use unprivileged overlayfs mounts whenever possible), so
supporting it in a more structured way should be very welcome.

One key thing to note is that (just like overlayfs) we only ever treat
one of "trusted.overlay." or "user.overlay." as special. The existing
tests also only needed miminal adjustments, and include tests to make
sure that we only ever touch the appropriate namespace.

This also finally allows us to run the overlayfs tests as an
unprivileged user.

Implements #576
Implements #584

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   (called `TranslateOverlayWhiteouts` in [umoci 0.4.7](#0.4.7)) now support
   converting `trusted.overlay.opaque=y` and `trusted.overlay.whiteout`
   whiteouts into OCI whiteouts when generating OCI layers.
+- `OverlayfsRootfs` now supports compatibility with the `userxattr` mount
+  option for overlayfs (where `user.overlay.*` xattrs are used rather than
+  the default `trusted.overlay.*`). This is a pretty key compatibility feature
+  for users that use unprivileged overlayfs mounts and will hopefully remove
+  the need for most downstream forks hacking in this functionality (such as
+  stacker). For Go API users, to enable this just set `UserXattr: true` in
+  `OverlayfsRootfs`. Note that (as with upstream overlayfs), only one xattr
+  namespace is ever used (so if `OverlayfsRootfs.UserXattr == true` then
+  `trusted.overlay.*` xattrs will be treated like any other non-overlayfs
+  xattr).
 
 ### Changes ###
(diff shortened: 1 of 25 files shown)


cyphar:
/cc @hallyn @tych0 @rchincha

This should remove the main need you have for a downstream fork of umoci. I suspect you will want to just do `OverlayfsRootfs{UserXattr: true}` for everything.

Issue #361, opened 2021-02-22
migrate "oci-image-tool validate" to umoci


cyphar:
This is the last step necessary to complete umoci's inclusion in the OCI -- to move validation into a more widely-known project so it can be used by more people. This does have the slight downside that the validation and reference implementation of the image-spec will live in the same repository but I don't think there's a much nicer solution (barring maintenance of image-tools to be revitalised).

cyphar:
This is now something we have to do because image-spec changes have made the old `oci-image-tool` runs stop working, and it's probably easier to just sit down and implement this. I took a quick look and it's not actually that bad (compared to `oci-runtime-tool`, we basically just have to use `jsonschema`...).

PR #601, opened 2025-07-28
oci: add SOURCE_DATE_EPOCH support for reproducible builds

danishprakash:
In order to make it easier for users to get reproducible builds out of
umoci without worrying about timestamp non-determinism, add support for
[SOURCE_DATE_EPOCH][1]. This affects most commands that involve adding
history entries or inserting layers.

For commands that just insert history entries or fill timestamp values,
SOURCE_DATE_EPOCH acts as the new default for those time.Time fields.
(But user-specified values still take precedence.)

For commands that insert layers, this patch implements "tar
--clamp-mtime" behavior (which is recommended by the spec), meaning that
only timestamps newer than SOURCE_DATE_EPOCH get modified and replaced
with SOURCE_DATE_EPOCH. Older timestamps are unaffected.

[1]: https://reproducible-builds.org/specs/source-date-epoch/

Signed-off-by: Danish Prakash <contact@danishpraka.sh>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,27 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added ###
 * `umoci stat` now includes information about the manifest and configuration of
   the image, both in the regular and JSON-formatted outputs.
+* umoci now has [`SOURCE_DATE_EPOCH`][source-date-epoch] support, to attempt to
+  make it easier to create reproducible images. Our behaviour is modelled after
+  `tar --clamp-mtime`, meaning that `SOURCE_DATE_EPOCH` will only be used to
+  modify the timestamps of files **newer** than `SOURCE_DATE_EPOCH`.
+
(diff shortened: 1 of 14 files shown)


cyphar:
There are some lint errors, but also the main CI is failing because `oci-image-tools validate` finally broke (I need to fix #361 first).

cyphar:
I fixed the CI issues a week or two ago, so this can be rebased now.

cyphar:
I'll do another rebase -- the `opensuse/leap:15.6` image just removed `file` which caused the tests to fail again.

cyphar:
The integration tests need a little more work, but I've also noticed that this is a good opportunity to finally add more information to `umoci stat`. I'm going to prepare a PR for that and rewrite the integration tests (and add some more) before merging.

cyphar:
By the way, I found a source of inconsistency when working on #624 -- gzip has an `mtime` timestamp in its header but we do not set it when generating gzip archives.

PR #645, opened 2025-10-30
Do not assume Linux OCI images (can be useful for FreeBSD, etc.)

nohajc:
I wanted to unpack a FreeBSD OCI image and had to remove the `PlatformOS` check first. Would it be OK to drop it altogether?


cyphar:
You can't just remove the check because that code generates an OCI runtime-spec `config.json` and it currently can only generate Linux-specific ones. [We just merged FreeBSD support into the spec](https://github.com/opencontainers/runtime-spec/pull/1286), so if you'd like FreeBSD support I would suggest the following course of action:

 1. Split the logic into per-OS files with `//go:build`.
 2. Add code to generate FreeBSD `config.json`s.
 3. Figure out how to run the CI tests on FreeBSD, and fix any issues that come up.

PR #675, opened 2026-05-03
feat(raw-unpack): Add --start-from-digest flag

kevinh-canva:
# Context

Add a small flag to `raw unpack` command to support unpacking from a specific layer.

Let me know if we want users to provide just the digest without `sha256:` prefix instead

# Demo

image


kevinh-canva:
Hi @cyphar / @tych0 - would either of you have some time to review this tiny PR? TIA

